PR #4, opened 2022-08-05
feat: Add Tooltip component and storybook

eric-zero-dev:
Screen Shot 2022-08-05 at 11 34 06 PM


src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  content?: ReactNode;
+  rootProps?: Partial<Omit<TooltipPrimitive.TooltipProps, 'open' | 'defaultOpen' | 'onOpenChange'>>;
+  triggerProps?: Partial<Omit<TooltipPrimitive.TooltipTriggerProps, 'asChild'>>;
+  contentProps?: Partial<Omit<TooltipPrimitive.TooltipContentProps, 'side' | 'align'>>;
+  arrowProps?: Partial<TooltipPrimitive.TooltipArrowProps>;
+  classNames?: {
+    triggerClass?: string;
+    contentClass?: string;
+    arrowClass?: string;
+  };
```
colbr, line 26: In my opinion, this granularity is breaking the idea of how our design system should work. For example, devs should not be able to restyle the arrow, because the arrow should have the same design throughout the design system. Similarly, they shouldn't be able to change how the trigger works (through `triggerProps`) because the tooltip should work the same everywhere.

I think we should:
- Remove `classNames` and add `className: string` (although I'm not even sure if we'd need a `className` here at all)
- Remove the `xProps` props (`rootProps`, `triggerProps`, `contentProps`, `arrowProps`)

Happy for discussion on this one!

src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import './Tooltip.scss';
+
+export const BASE_CLASSNAME = `${CSS_PREFIX}-tooltip`;
+
+export type TooltipProps = {
+  open?: TooltipPrimitive.TooltipProps['open'];
+  defaultOpen?: TooltipPrimitive.TooltipProps['defaultOpen'];
+  onOpenChange?: TooltipPrimitive.TooltipProps['onOpenChange'];
+  side: TooltipPrimitive.TooltipContentProps['side'];
+  align: TooltipPrimitive.TooltipContentProps['align'];
```
colbr, line 16: Liking this pattern of pulling types straight from Radix - nice!

src/components/constants.ts
```diff
@@ -0,0 +1 @@
+export const CSS_PREFIX = 'ZUI';
```
colbr, line 1: Can we make this lowercase?

src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,78 @@
+import type { FC, ReactNode } from 'react';
```
colbr, line 1: I think I'm leaning towards putting all components in a title case file (i.e. `Tooltip.tsx`) and just using `index.ts` for re-exporting. Makes the DX a little bit nicer.

colbr (commented):
Looking good! I suggested a couple of things - let me know what you think.

Ping me in Discord when you've responded so I know to come back and check.

colbr (approved):
Nice one!

joonchoi0918 (commented):
Looks good to me. Can you add test for tooltip?

colbr:
> Looks good to me. Can you add test for tooltip?

Agreed, this would be handy

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    expect(getByTestId(TOOLTIP_INNER_DIV_ID)).toBeTruthy();
+  });
+
+  it('Should correctly sets the content attribute of the Tooltip component', () => {
+    const { queryAllByText } = renderComponent(true);
+
+    expect(queryAllByText(TOOLTIP_CONTENT)).toBeTruthy();
+  });
+});
```
colbr, line 47: Can we change this to use `react-testing-library` syntax, e.g. `test('correctly renders the trigger attribute')`?

nit: I don't think we need the "of the tooltip component" in each test name as the test runner prints the name of the component before running the suite.

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
colbr, line 1: I think we're missing some tests here:
`applies open prop to Radix tooltip root`, `applies defaultOpen prop to Radix tooltip root`, `applies onOpenChange prop to Radix tooltip root`, `applies side prop to Radix tooltip content`, `applies align prop to Radix tooltip content`.

Bit of a weird set of tests because generally we wouldn't write tests for implementation details, but for these tests, we want to know that we're configuring the Radix primitives correctly.

colbr (changes requested):
A few small changes, but looking good overall 🔥

PR #11, opened 2022-08-18
feat: add grid view component to async table

domw30:
Edit Async Table to enable gridComponent


.gitignore
```diff
@@ -6,4 +6,4 @@ dist/
 
 .npmrc
 *.tgz
-yarn-error.log
\ No newline at end of file
+yarn-error.log
```
domw30, line 9: hmm this shouldn't be an addition on my PR. Is this required in gitignore?

PR #23, opened 2022-08-26
refactor: remove default exports

colbr:
We have decided to [remove default exports](https://docs.google.com/document/d/1y1zRw0-4bFaS9Lw3MLowMd248g2QrzV9i0Pq1iG4HkU/edit?usp=sharing).

- Added an ESLint rule to prevent default exports (`"import/no-default-export": "off"`).
- Removed default exports from all components.
- Changed stories and tests to use named imports rather than default.


src/components/Modal/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Modal } from './Modal';
+export type { ModalProps } from './Modal';
```
joonchoi0918, line 2: We have 2 exports `ModalProps ` and `Modal`, but the `ModalProps` is not used in components expose. And there are only 2 exports in `./Modal`. So we can use it `export * from './Modal';`

src/components/index.ts
```diff
@@ -1,12 +1,12 @@
...
-export { default as LoadingIndicator } from './LoadingIndicator';
-export { default as Modal } from './Modal';
-export { default as Skeleton } from './Skeleton';
+export { AsyncTable } from './AsyncTable';
+export { Button } from './Button';
+export { Card } from './Card';
+export { DropdownMenu } from './DropdownMenu';
+export { Input } from './Input';
+export { LoadingIndicator } from './LoadingIndicator';
+export { Modal } from './Modal';
```
joonchoi0918, line 7: We exported only `Modal` here. I am pretty sure we can use `export * from './Modal';` here as well. Then we can use `import { Modal } from '@zero-tech/zUI/components';`  and `import type { ModalProps } from '@zero-tech/zUI/components';`

src/components/Modal/index.ts
eric-zero-dev: I agree with @joonchoi0918

src/components/index.ts
eric-zero-dev: Yeah, Exactly! We could use `ModalProps` in zApps components for type definition or some type references.

src/components/Modal/index.ts
colbr: I am personally not a huge fan of `export * from './Modal'` - I like to keep exports very explicit. Although, `export *` may be fine since someone could just do `import * from 'zui/Modal/Modal` and skip the index anyway 🤷‍♂️ Will see what @jtulloch says.

src/components/index.ts
colbr: Yeah good point, we could use `export * from './Modal'` here. Will wait on @jtulloch to review before pulling the trigger.

src/components/Modal/index.ts
colbr: On second thought, yeah I think `export *` is ok here. My initial hesitation with them is we would have to be careful with what we export from `Modal`, but with either method we have to be careful with one or the other so it's arbitrary.

jtulloch: agreed. i think it's fine here.

src/components/index.ts
jtulloch: yeah. i think it's worth being a *bit* careful what gets propagated into the package interface as a whole, and would lean more towards what @colbr said in the earlier comment about making exports explicit. but at the same time. .. probably not a huge deal, really, as i doubt we'll be exporting much at the component level that we wouldn't want at the package level.

colbr: Awesome thanks for the input guys - I'll use `export *` for these and merge. Cheers!

colbr:
Changed the commit type to `refactor`

PR #24, opened 2022-08-26
test(card): improve testing

colbr:
- Added test for order of card elements.
- Added tests for synchronous and asynchronous `value` & `bottomText`.


colbr:
Went through a few iterations with this one, but got there in the end.

jtulloch (changes requested):

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,97 @@
...
+  const asText = Array.from(children).map(c => c.textContent);
+  const asNodes = Array.from(children).map(c => c.nodeName);
+
+  // Assert correct nodes and correct order
+  const expectedNodes = ['label', 'b', 'span'];
+  const expectedText = [DEFAULT_PROPS.title, DEFAULT_PROPS.value, DEFAULT_PROPS.bottomText];
+  expect(
+    asText.toString() === expectedText.toString() &&
+      asNodes.map(n => n.toLowerCase()).toString() === expectedNodes.map(n => n.toLowerCase()).toString()
+  );
```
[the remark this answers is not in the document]
colbr: I felt like it was a little draconian while writing it - nice to hear that you also agree. I'll remove this one.

I think I'd like to assert the `<label>` node type as it's important for screen readers, but the other assertions can go for sure.

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,97 @@
...
+});
+
+test('should apply class module names to correct elements', () => {
+  const { getByText } = renderComponent();
+
+  expect(getByText(DEFAULT_PROPS.title)).toHaveClass(styles.Title);
+  expect(getByText(DEFAULT_PROPS.value)).toHaveClass(styles.Value);
+  expect(getByText(DEFAULT_PROPS.bottomText)).toHaveClass(styles.BottomText);
+});
```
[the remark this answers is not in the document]
colbr: Sweet - have handled most of these. Agreed on making smaller tests.

I'm finding it difficult to remove `DEFAULT_PROPS`. I'm using it here because, if I didn't, I'd have to repeatedly define the props per test. If I added a new required prop to `Card`, I'd have to go through each test and add it, which adds friction.

Would `const { ... } = renderComponent(DEFAULT_PROPS)` be an acceptable middle-ground? This is more explicit than handling them in the `renderComponent` function, and lets me reduce repetition.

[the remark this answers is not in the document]
jtulloch: @colbr maybe my comment just wasn't clear enough. my suggestion still solves the problem you're referring to. you shouldn't be defining all of the props in every test. what i am referring to is that you *should* be explicitly providing the props that are important to each given test. if you think it makes it more explicit to provide the defaults in each test, then maybe go for something like this:
`const { ... } = renderComponent(withDefaults({ title: 'my title!' }));`

(i will also add that setting sane defaults for every test, which happens if you have a global setup, actually hinders the scope of what you're testing. using the bare minimum for defaults (nulls for objects, empty strings, zero, etc.) actually helps make sure that your component handles situations where data isn't fully formed yet.)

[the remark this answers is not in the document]
colbr: @jtulloch ohhh ok I think I read this wrong initially. Just to be clear - you're cool with using `DEFAULT_PROPS` in `renderComponent`, and then overriding `DEFAULT_PROPS` with whatever is passed to `renderComponent`? e.g.

```js
// psuedo
const renderComponent  = (props) => {
    return render(<Component {...props} {...DEFAULT_PROPS} />)
}

// in test
const { ... } = renderComponent({ title: 'mock title' });
```

I think I would prefer to use `const { ... } = render(<Component title="mock title" {...DEFAULT_PROPS} />)` the below if you think it's an acceptable option.

> i will also add that setting sane defaults for every test...

Is this relevant to these tests? I'm not really sure what the benefit of setting minimum defaults would be if we were passing all relevant props per test. I'm only really using `DEFAULT_PROPS` to satisfy types. I would think the non-sane values like `''` and `null` would be handled in tests. If possible could you provide an example from this test suite? I'm not quite sure I'm following and I think a concrete example would help.

Although now that I think about it, we haven't tested `bottomText is undefined`.

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+const MOCK_SKELETON_ID = 'mock-skeleton';
+
 jest.mock('../Skeleton', () => ({
-  Skeleton: () => <span>Skeleton</span>
+  Skeleton: () => <div data-testid={MOCK_SKELETON_ID} />
 }));
 
-const renderComponent = (props?: CardProps) => render(<Card {...props} />);
+describe('<Card />', () => {
+  test('should only render text in the correct order - title, value, bottom text', () => {
```
domw30, line 21: The way I read this test suggests a specific order of ```1,2,3```, _only_. Might be a little overkill(?), but should checks be added to confirm text content won't be in an unexpected order, something like ```expect(container).not.toHaveTextContent('3,2,1')```? 

Although, I guess this is part of the reason you've added commas to ensure an exact match because without them something like ```expect(container).not.toHaveTextContent('321')``` should still pass.

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+    const { container } = render(<Card {...DEFAULT_PROPS} title={'1,'} value={'2,'} bottomText={'3'} />);
```
domw30, line 22: is ```{...DEFAULT_PROPS}``` required if defining props here?

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
-});
+  describe('class names', () => {
+    test('should apply Title class to only the title element', () => {
+      const { container } = render(<Card {...DEFAULT_PROPS} title={'mock title'} />);
+      const title = container.getElementsByClassName(styles.Title);
+      expect(title.length).toBe(1);
+      expect(title[0]).toHaveTextContent('mock title');
+    });
+
+    test('should only apply Value class to only the value element', () => {
```
domw30, line 43: Should this check other classes aren't applied? Its just the way I perceive the use of word _only_. Same with other tests.

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+      const value = container.getElementsByClassName(styles.Value);
+      expect(value.length).toBe(1);
+      expect(value[0]).toHaveTextContent('mock value');
+    });
+
+    test('should only apply BottomText class to the bottom text element', () => {
+      const { container } = render(<Card {...DEFAULT_PROPS} bottomText={'mock bottom text'} />);
+      const bottomText = container.getElementsByClassName(styles.BottomText);
+      expect(bottomText.length).toBe(1);
+      expect(bottomText[0]).toHaveTextContent('mock bottom text');
```
domw30, line 54: As we're only testing ```BottomText``` class is applied, would it make more sense to reduce this test to something like:

```
test('should only apply BottomText class to the bottom text element', () => {
  const { container } = render(<Card {...DEFAULT_PROPS} bottomText={'mock bottom text'} />);
  expect(container.getElementsByClassName(styles.BottomText)).toBeTruthy();
});
```

I'd say we could separate the expects for elements length and text content in this and the above tests if our aim is to ensure each test is scoped to exactly what it asks.

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+    });
+  });
+
+  describe('when text props are synchronous', () => {
+    test('should render title', () => {
+      render(<Card {...DEFAULT_PROPS} title={'mock title'} />);
+      expect(screen.getByText('mock title')).toBeInTheDocument();
```
domw30, line 61: screen or 
```
const { getByText } = render(<Card {...DEFAULT_PROPS} bottomText={'mock bottom text'} />);
```

Not fussed myself, but we should move forward with a consistent rule

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+    });
+
+    test('should render value in the correct element (<b>)', () => {
```
domw30, line 64: If the intention is to check value is rendered in/with the correct element ```<b>```, should we use something like:
```
const value = container.getElementsByTagName('<b>');
expect(value).toHaveTextContent('mock value')
```

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
+    });
+
+    test('should render synchronous bottomText', () => {
+      render(<Card {...DEFAULT_PROPS} bottomText={'mock bottom text'} />);
+      expect(screen.getByText('mock bottom text')).toBeInTheDocument();
+    });
+  });
+
+  describe('when bottom text is asynchronous', () => {
+    test('should show loading skeleton when loading bottom text', () => {
```
domw30, line 76: do you think in scenarios like this we should check that bottom text is falsy when loading? Not sure how far we want to take coverage

similar to below, would you expect to assert that skeleton is falsy when text/value is truthy

domw30 (commented):
Hey @colbr, dropped in a few questions but liking what I see for moving forward with testing

src/components/Card/Card.test.tsx
```diff
@@ -1,37 +1,98 @@
...
 
+const MOCK_SKELETON_ID = 'mock-skeleton';
+
 jest.mock('../Skeleton', () => ({
-  Skeleton: () => <span>Skeleton</span>
+  Skeleton: () => <div data-testid={MOCK_SKELETON_ID} />
 }));
 
-const renderComponent = (props?: CardProps) => render(<Card {...props} />);
+describe('<Card />', () => {
```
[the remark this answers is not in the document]
colbr: Perfect!

src/components/Card/Card.test.tsx
colbr: This test here is here to assert that the visible elements are rendered in the order `title`, `value`, `bottom text`. I think my use of `only` might have made this test a bit confusing, but in essence it's checking that we're seeing:
```
_Title_
**Value**
_Bottom Text_
```
Rather than some other variation of the above, e.g:
```
**Value**
_Title_
_Bottom Text_
```
My thinking is that we can assert that this is true by using three arbitrary strings as props and checking that the text content of the container is equal to those three strings concatenated in the right order. If we had `title=1`, `value=2`, and `bottomText=3`, and the container text content was `123`, we would know that the text ordering is correct.

I put commas in each of the strings just so the test output reads a little nicer - they're not important to the test. Perhaps this was a little bit confusing and should be changed.

I think you're right that I need to use something stricter than `toHaveTextContent` though as it's not an exact match. I will change this to `expect(container.textContent).toBe('1,2,3')`.

src/components/Card/Card.test.tsx
colbr: Nope - thanks for spotting that.

src/components/Card/Card.test.tsx
colbr: Hmm yes perhaps `only` isn't the right word to use here. I was checking that `styles.Value` is applied to the correct element and no others.

I just realised that there's a bit of a typo in here with the double `only` - it should be `should apply Value class to only the value element`.

src/components/Card/Card.test.tsx
colbr: It's tricky - the suggested checks that the class was applied to an element, but not the correct element. I found this a pretty tough thing to test without using `testId` - perhaps this is a situation where `testId` works well!

e.g.
```
render(...);
const bottomText = screen.getByTestId(...);
expect(bottomText).toHaveClass(...);
```

src/components/Card/Card.test.tsx
colbr: We should [use screen](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen).

> The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.
> The only exception to this is if you're setting the container or baseElement which you probably should avoid doing (I honestly can't think of a legitimate use case for those options anymore and they only exist for historical reasons at this point).
> You can also call [screen.debug](https://testing-library.com/docs/dom-testing-library/api-queries#screendebug) instead of debug

src/components/Card/Card.test.tsx
colbr: Oops, forgot to update the test name. Thanks for pointing this out!

src/components/Card/Card.test.tsx
colbr: Yeah probably worth checking the Skeleton is no longer in the DOM - I'll add it to the tests.

src/components/Card/Card.test.tsx
colbr: Note: I know this one says not to destructure `{ container }` but I felt it was necessary on some of the tests. I would prefer not to add a test ID to the container if we can get the container directly from the `render` function.

src/components/Card/Card.test.tsx
domw30: Yeah nice I like this - ```expect(container.textContent).toBe('1,2,3')```. I copied the above for ordering of TabsNav but I'll go back and change it to this, makes sense

src/components/Card/Card.test.tsx
domw30: Ahhh no, I see what you mean. Checking the length covers the 'only' part of the scope. Probably don't need to check the text content, unless that hasn't been covered in another test, in that case I'd leave this as is 😁

src/components/Card/Card.test.tsx
domw30: Sweet, sounds good to me!!

PR #28, opened 2022-09-05
feat: add hooks and format utility in storybook

joonchoi0918:
image


src/components/Utils/Utils.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <ul>
+        {utils.map(hook => (
+          <li key={hook} className="utils-stories__list-item">
+            {hook}
+          </li>
+        ))}
+      </ul>
+    </StoryCard>
+  ));
+};
```
colbr, line 42: I don't think this is the best way for us to show these functions, as it doesn't provide much more information than "these functions exist". Non UI elements shouldn't have UI stories - we should [use MDX](https://stackoverflow.com/questions/70365222/is-it-possible-to-write-doc-for-function-class-non-ui-elements-in-storybook) instead.

What would be ideal, is if we split them up into sections. For example, we should have a `Formatting` section under `Utils`, which has the `format...` functions.

For each section, we should list each function and what its output looks like.

src/utils/format.ts
```diff
@@ -0,0 +1,25 @@
+import { BigNumber } from 'ethers';
```
colbr, line 1: Can we bring over the `format.ts` tests as well?

src/utils/hooks/useDidMount.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, EffectCallback } from 'react';
```
colbr, line 1: Remove this hook.

src/utils/hooks/useDidUpdate.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef, EffectCallback, DependencyList } from 'react';
```
colbr, line 1: Remove this hook.

src/utils/hooks/usePrevious.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useRef } from 'react';
```
colbr, line 1: Remove this hook.

src/utils/hooks/usePropsState.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
```
colbr, line 1: Remove this hook.

src/utils/hooks/useWillUnmount.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react';
```
colbr, line 1: Remove this hook.

colbr (changes requested):
Good start.

Like I said in the comments, we should use MDX for non-UI stories.

I think we should remove all of the hooks as well, as they're not directly related to the UI. We should keep zUI for code that directly affects the UI - formatting utilities are a good example of this as they alter some text which is rendered.

We can think about adding some utility hooks at a later date, but I think I'd rather just have them in a zApp Utility package.

PR #29, opened 2022-09-07
refactor(modal): remove CSS-in-JS [skip ci]

colbr:
- Removes old CSS-in-JS styling from the `Modal` component.
- Adds `Modal.module.scss` for styling

### DIFF
--- a/src/components/Modal/Modal.module.scss
+++ b/src/components/Modal/Modal.module.scss
@@ -16,9 +16,9 @@
 }
 
 .Content {
-  min-width: 300px;
-  padding: 30px;
-  border-radius: 16px;
+  min-width: 18.75rem; // 300px
+  padding: 1.875rem; // 30px
+  border-radius: 1rem;
   border: 1px solid #{$grey-dark};
 
   background: #{$almost-black};
--- a/src/components/Modal/Modal.module.scss
+++ b/src/components/Modal/Modal.module.scss
@@ -1,0 +1,23 @@
+.Overlay {
+    display: grid;
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+
+    place-items: center;
+    overflow-y: auto;
+    z-index: 10000;
+
+    background: rgba(0 0 0 / 0.9);
+}
+
+.Content {
+    min-width: 300px;
+    padding: 30px;
+    border-radius: 16px;
+    border: 1px solid #3b3b3b;
+
+    background: #0a0a0a;
+}
--- a/src/components/Modal/Modal.module.scss
+++ b/src/components/Modal/Modal.module.scss
@@ -1,23 +1,25 @@
+@import '../../styles/colors';
+
 .Overlay {
-    display: grid;
-    position: fixed;
-    top: 0;
-    left: 0;
-    right: 0;
-    bottom: 0;
+  display: grid;
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
 
-    place-items: center;
-    overflow-y: auto;
-    z-index: 10000;
+  place-items: center;
+  overflow-y: auto;
+  z-index: 10000;
 
-    background: rgba(0, 0, 0, 0.9);
+  background: rgba(0, 0, 0, 0.9);
 }
 
 .Content {
-    min-width: 300px;
-    padding: 30px;
-    border-radius: 16px;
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -3653,6 +3653,11 @@
     "@testing-library/dom" "^8.0.0"
     "@types/react-dom" "<18.0.0"
 
+"@testing-library/user-event@^14.4.3":
+  version "14.4.3"
+  resolved "https://registry.yarnpkg.com/@testing-library/user-event/-/user-event-14.4.3.tgz#af975e367743fa91989cd666666aec31a8f50591"
+  integrity sha512-kCUc5MEwaEMakkO5x7aoD+DLi02ehmEM2QCGWvNqAS1dV/fAvORWEjnjsEIvml59M7Y5kCkWN6fCCyPOe8OL6Q==
+
```
joonchoi0918, line 3660: You got this issue. The another developer didn't update it in his PR 😄

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    background: rgba(0, 0, 0, 0.9);
+}
+
+.Content {
+    min-width: 300px;
+    padding: 30px;
+    border-radius: 16px;
+    border: 1px solid #3b3b3b;
+
+    background: #0a0a0a;
```
joonchoi0918, line 22: Don't we use the colors defined in `styles/_colors.scss`?

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,23 @@
...
+    display: grid;
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+
+    place-items: center;
+    overflow-y: auto;
+    z-index: 10000;
```
joonchoi0918, line 11: I recommend to use some variable for z-index something like `$z-index-100~10000` or  `$z-index-xxs~xxl`

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+    background: rgba(0, 0, 0, 0.9);
+}
+
+.Content {
+    min-width: 300px;
+    padding: 30px;
+    border-radius: 16px;
```
joonchoi0918, line 19: nit:  I think we moved to using `px` from `rem`

src/components/Modal/Modal.tsx
```diff
@@ -8,32 +8,11 @@ import {
...
-  zIndex: 10000
-});
-
-const Content = styled(DialogContent, {
-  minWidth: 300,
-  background: '#0a0a0a',
-  padding: 30,
-  borderRadius: 16,
-  border: '1px solid #3b3b3b'
-});
```
joonchoi0918, line 36: I really couldn't understand this styled component, but I am happy to see removing this 😄

joonchoi0918 (approved):
Great Work! I left a few comments about variables and colors. But I know we will get rid of these legacy colors when we use `ThemeColors` and I think it is fine for now. Not sure you used the color which is in the color pallets.

src/components/Modal/Modal.tsx
colbr: I was just testing out Radix when I made this component - they used Stitches in the example which I more or less copied, haha. I'm glad to purge it too!

yarn.lock
colbr: Probably me, haha 🤡 All good, should be sorted once this is merged.

src/components/Modal/Modal.module.scss
colbr: Good point, let me fix that

src/components/Modal/Modal.module.scss
colbr: I think in this case we wouldn't want anything to share the same zIndex as the `Modal` component, as it's meant to be on top. Let me know if this assumption is incorrect though!

src/components/Modal/Modal.module.scss
colbr: Good call - I'll switch them.

src/components/Modal/Modal.module.scss
colbr: Fixed!

src/components/Modal/Modal.module.scss
colbr: Fixed!

PR #31, opened 2022-09-07
chore: configure Husky pre-commit lint + prettify

colbr:
This should (hopefully) improve DX by making linting automatic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,11 @@
     "test": "jest",
     "test:watch": "jest --watch"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn lint"
+    }
+  },
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -20,6 +20,11 @@
     "test": "jest",
     "test:watch": "jest --watch"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn lint"
+    }
+  },
```
joonchoi0918, line 27: Great!

joonchoi0918: nit: Can we add `prettify` command as well? So prettier first and lint sencond.
```
"husky": {
    "hooks": {
      "pre-commit": ["yarn prettify", "yarn lint"]
    }
  },
```

joonchoi0918 (approved):
Nice work! Thanks 🙏

package.json
colbr: Sure thing - will add!

PR #34, opened 2022-09-14
test(dropdown-menu): increase coverage

colbr:
- Increased testing coverage of DropdownMenu.
- Refactored DropdownMenu slightly.

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.tsx
@@ -13,40 +13,48 @@ import './DropdownMenu.scss';
 import classNames from 'classnames';
 
 export interface DropdownItem {
+  /** Unique ID to use in Array.map() */
   id: string;
+  /** Label to appear in the menu */
   label: string | ReactNode;
+  /** Callback to run when an item is selected */
   onSelect: (event?: Event) => void;
   className?: string;
-  icon?: string | ReactNode;
 }
 
 export interface DropdownMenuProps extends RadixUIDropdownMenuProps {
+  /** List of items to render in the dropdown menu */
   items: DropdownItem[];
-  trigger?: ReactNode | 'triple-dot';
+  /** Clicking this will make the dropdown menu appear */
+  trigger?: ReactNode;
 
   // from RadixDropdownMenu.DropdownMenuContentProps
   side?: 'top' | 'right' | 'bottom' | 'left';
   alignMenu?: 'start' | 'center' | 'end';
 
+  /** Class to apply to the trigger */
   className?: string;
+  /** Class to apply to the dropdown menu */
   menuClassName?: string;
 }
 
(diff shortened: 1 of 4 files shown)


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -13,40 +13,46 @@ import './DropdownMenu.scss';
 import classNames from 'classnames';
 
 export interface DropdownItem {
+  /** Unique ID to use in Array.map() */
   id: string;
+  /** Label to appear in the menu */
   label: string | ReactNode;
+  /** Callback to run when an item is selected */
```
jtulloch, line 20: do we need these comments? they don't really seem to provide much benefit.

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -13,40 +13,46 @@ import './DropdownMenu.scss';
...
-  menuClassName?: string;
 }
 
 export const DropdownMenu = forwardRef<HTMLDivElement, DropdownMenuProps>(
-  ({ className, alignMenu = 'center', items, trigger = 'triple-dot', side = 'bottom', ...dropdownMenuProps }, ref) => {
+  ({ className, alignMenu = 'center', items, trigger, side = 'bottom', ...dropdownMenuProps }, ref) => {
+    const triggerElement = !trigger ? <span>...</span> : typeof trigger === 'string' ? <span>{trigger}</span> : trigger;
+
     return (
       <RadixUIDropdownMenuRoot {...dropdownMenuProps}>
```
jtulloch, line 44: as per our discussion in the chat room, let's not arbitrarily forward props. it's slightly more convenient initially, but the risks are just too great. aside from the risk factor. .. being explicit encourages thinking about what things are *actually* important to the current component, and can also be a signal when a component is getting too complex. it also helps reduce over-typing, and helps avoid tying your interface too tightly to another components implementation.

jtulloch (changes requested):
tests look good. just two comments.

src/components/DropdownMenu/DropdownMenu.tsx
colbr: The intention was to improve DX as per below:
Screen Shot 2022-09-21 at 1 47 03 PM

However, looks like these comments aren't included in the build. That puts a damper on things!

src/components/DropdownMenu/DropdownMenu.tsx
colbr: On it.

colbr:
Handled your feedback @jtulloch

src/components/DropdownMenu/DropdownMenu.tsx
jtulloch: ah gotcha. .. :thinking: yeah. .. that makes sense. if you can figure out how make that work, i agree that it would be pretty useful in this type of situation.

PR #36, opened 2022-09-17
feat: add Link component

colbr:


src/components/Link/Link.types.ts
```diff
@@ -0,0 +1,20 @@
+import { ReactNode } from 'react';
```
eric-zero-dev, line 1: nit: `import type { ReactNode }  from 'react';`

src/components/Link/RouterLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { ComponentProps } from './Link.types';
+import { Link as ReactRouterLink } from 'react-router-dom';
+
+import classNames from 'classnames';
+import styles from './Link.module.scss';
+
+export interface RouterLinkProps extends ComponentProps {
+  to: string;
+  replace?: boolean;
```
eric-zero-dev, line 11: nit: We can use the type from LinkProps

```
import type { LinkProps } from 'react-router-dom';

export interface RouterLinkProps extends ComponentProps {
  to: LinkProps['to'];
  replace: LinkProps['replace'];
}
```

This is the LinkProps definition - https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L341-L349

src/components/Link/RouterLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const RouterLink = ({ to, replace, children, className }: RouterLinkProps) => {
+  console.log('got classname:', className);
```
eric-zero-dev, line 15: console.log?

eric-zero-dev (changes requested):
I left a few comments with minimal type imports. Looks good to me.

src/components/Link/RouterLink.tsx
eric-zero-dev: We might need to pass `pathname` and `state` into `to` props.

src/components/Link/BackLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+/**
+ * Navigates back using react-router-dom useNavigate.
+ * Override back behaviour by passing onClick.
+ */
+export const BackLink = ({ onClick, ...rest }: BackLinkProps) => {
+  const navigate = useNavigate();
+
+  const goBack = () => navigate(-1);
```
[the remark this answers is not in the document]
domw30: I think the behaviour would be the same as that of the ```<- My Profile``` button in zns-dApp. So in that scenario the user would be redirected to the site they were on previously. Agreed though, may want to handle it if previous page is a different domain/external site

src/components/Link/ArrowLink/ArrowLink.stories.tsx
```diff
@@ -4,16 +4,16 @@ import { ArrowLink } from '.';
...
   return (
     <StoryCard isContrast>
-      <ArrowLink {...args} />
+      <ArrowLink {...args}>Arrow Link</ArrowLink>
     </StoryCard>
   );
 };
 
-export const Default = Template.bind({});
+export const Arrow = Template.bind({});
```
domw30, line 19: Looks good in storybook 👍

src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,37 @@
...
+        text-decoration: none;
+        outline: none;
+        border: none;
+        background: none;
+
+        display: inline-block;
+        width: fit-content;
+
+        // @TODO: update with theme engine variable
+        color: white !important;
```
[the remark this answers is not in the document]
domw30: +1. Although these look like they've been added to a Todo list 😁

src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,37 @@
...
+
+    &.Primary {
+        text-decoration: none;
+
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+
+    &.Secondary {
```
domw30, line 27: is the ```&``` required on these nested styles?

domw30 (approved):
Looks good 👍

src/components/Link/BackLink.tsx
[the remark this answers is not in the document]
colbr: Oh yeah, good catch. I don't think there's actually going to be a nice way to make a generic back button, so perhaps we should omit it.

src/components/Link/Link.module.scss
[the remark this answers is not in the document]
colbr: Yep yep, just leaving as white for now as I imagine ThemeEngine will be merged before this. Good find though, thanks for commenting on it.

src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,37 @@
...
+        outline: none;
+        border: none;
+        background: none;
+
+        display: inline-block;
+        width: fit-content;
+
+        // @TODO: update with theme engine variable
+        color: white !important;
+        font-size: 1rem;
```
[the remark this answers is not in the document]
colbr: I think we should have a team chat about how we're going to handle font - I'll chat in Discord.

src/components/Link/RouterLink.tsx
colbr: Good call!

src/components/Link/RouterLink.tsx
colbr: Thanks for picking it up - leaving it in for now while I work on link but will make sure to delete before submitting again.

src/components/Link/Link.module.scss
colbr: For this one, `.Secondary` is added to the same element as `.Container`, so need the `&`.

PR #38, opened 2022-09-21
test(loading-indicator): improve testing

domw30:
- Removed use of testIds as per initial zUI code review notes.
- Refactor tests to follow existing patterns (card/dropdown)

### DIFF
--- a/src/components/LoadingIndicator/LoadingIndicator.test.tsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.test.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
-import { cleanup, render } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { LoadingIndicator } from './LoadingIndicator';
 
 enum MockSpinnerPositionType {
@@ -11,7 +11,7 @@ enum MockSpinnerPositionType {
 
 const MOCK_DEFAULT_PROPS = {
   className: 'mock-class-name',
-  text: 'Mock Text',
+  message: 'mock message',
   spinnerPosition: MockSpinnerPositionType.LEFT
 };
 
@@ -21,29 +21,14 @@ jest.mock('./Spinner', () => ({
   Spinner: () => <div>{MOCK_SPINNER}</div>
 }));
 
-// poss not needed
-afterEach(() => {
-  cleanup();
-  jest.clearAllMocks();
-});
-
-describe('LoadingIndicator', () => {
+describe('<LoadingIndicator />', () => {
+  // defined
   test('should render loading indicator container', () => {
     const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
+    expect(container.firstChild).toHaveClass('zui-loading-indicator');
+
     const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
-
-    // Imp// feedback testrove this test
-    expect(container.firstElementChild).toBeTruthy();
-
     expect(loadingIndicatorContainer.length).toBe(1);
-  });
-
-  test('should render loading indicator text container and text value', () => {
-    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
-    const textContainer = container.getElementsByClassName('zui-loading-indicator-text');
-
-    expect(textContainer.length).toBe(1);
-    expect(textContainer[0].textContent).toEqual(MOCK_DEFAULT_PROPS.text);
   });
 
   test('should render spinner component', () => {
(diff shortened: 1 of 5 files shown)


domw30:
what should we change name of text too, as Joel suggested? Does the text type need to include ReactNode ? 🤔

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
+/* eslint-disable no-unused-vars */
 import React from 'react';
 import { render } from '@testing-library/react';
-import { LoadingIndicator, LoadingIndicatorProps, TEST_ID } from './LoadingIndicator';
+import { LoadingIndicator, LoadingIndicatorProps } from './LoadingIndicator';
 
-const renderComponent = (props: LoadingIndicatorProps) => render(<LoadingIndicator {...props} />);
+enum MockSpinnerPositionType {
```
domw30, line 6: Not sure why this was causing linting warning but likely I'm missing something

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
+    expect(textContainer[0].textContent).toEqual(MOCK_DEFAULT_PROPS.text);
+  });
+
+  test('should render spinner component', () => {
+    const { getByText } = renderComponent();
+
+    expect(getByText(MOCK_SPINNER)).toBeTruthy();
+  });
+
+  describe('when props are defined', () => {
```
domw30, line 56: These describe's could maybe be clearer if you like? Noticing others don't really use these, happy to remove if you want, I quite like them in the logs when running tests

colbr:
> what should we change name of text too, as Joel suggested? Does the text type need to include ReactNode ? 🤔

[`string` is actually a subset of `ReactNode`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2034c45/types/react/index.d.ts#L203), so including both is redundant. I think we should probably make this one `ReactNode` since it's not a Figma component (it's just a utility component for us to use), in which case `text` might not be the best prop name.

src/components/LoadingIndicator/LoadingIndicator.test.tsx
colbr: `describe` is a handy tool, because you can create scoped context which can reduce duplicated code, e.g.
```
describe('something', () => {
    beforeEach(() => {
        // do some setup here that all tests use
    })

    test('something related', () => {
        // use context created by local beforeEach
    })
})
```
`describe` also just creates better signposting - easier to find the test you're looking for if they're grouped nicely. We should probably be using `describe` more often.

I think the `describe`s in this test could be improved. Perhaps grouping them by what feature is being tested would be cleaner - `describe('spinner positioning', ...)`, `describe('class names', ...)`. If you have a look at [the Jest docs here](https://jestjs.io/docs/setup-teardown#scoping), they have `describe('matching cities to foods', ...)` which is a nice high level description of the feature they're testing.

colbr (changes requested):
Awesome these are solid. Thanks for picking these up.

Left some comments, I think we're pretty close to having a good standard for component tests.

One more test to add - `spinnerPosition` relies on the spinner being the first element in the container. It uses flex `reverse` properties to set positioning, so if the spinner isn't first the `spinnerPosition` will be reversed. Perhaps we could make this easier to test if the markup changed for different `spinnerPosition`s, rather than a class name.

Nice work all round though, looking forward to seeing the next iteration.

src/components/LoadingIndicator/LoadingIndicator.test.tsx
domw30: 'MockSpinnerPositionType' is defined but never used.

no unused vars warning. added /* eslint-disable no-unused-vars */. Will try figure out why thats being raised, shouldn't be

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
-});
+const renderComponent = (props?: LoadingIndicatorProps) =>
+  render(
+    <LoadingIndicator
+      className={MOCK_DEFAULT_PROPS.className}
+      text={MOCK_DEFAULT_PROPS.text}
+      spinnerPosition={MOCK_DEFAULT_PROPS.spinnerPosition}
+      {...props}
+    />
+  );
```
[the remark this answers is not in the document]
domw30: sounds good to me. Updated!

src/components/LoadingIndicator/LoadingIndicator.test.tsx
domw30: Sweet thanks will check this out and update

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
 
-test('should render spinner', () => {
-  const { getByTestId } = renderComponent({});
-  const spinner = getByTestId(TEST_ID.SPINNER);
-  expect(spinner).toBeInTheDocument();
-});
+describe('LoadingIndicator', () => {
+  test('should render loading indicator container', () => {
```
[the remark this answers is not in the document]
domw30: Done

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
+  });
+
+  test('should render spinner component', () => {
+    const { getByText } = renderComponent();
+
+    expect(getByText(MOCK_SPINNER)).toBeTruthy();
+  });
+
+  describe('when props are defined', () => {
+    test('should apply a custom className to loading indicator container className attribute', () => {
```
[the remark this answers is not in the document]
domw30: Done :)

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
+      const { container } = renderComponent();
+      const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
+
+      expect(loadingIndicatorContainer[0]).toHaveClass(MOCK_DEFAULT_PROPS.className);
+    });
+
+    test('should append spinnerPosition type to loading indicator container className attribute', () => {
```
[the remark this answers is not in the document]
domw30: Done

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
+    expect(loadingIndicatorContainer.length).toBe(1);
+  });
+
+  test('should render loading indicator text container and text value', () => {
+    const { container } = renderComponent();
+    const textContainer = container.getElementsByClassName('zui-loading-indicator-text');
+
+    expect(textContainer.length).toBe(1);
+    expect(textContainer[0].textContent).toEqual(MOCK_DEFAULT_PROPS.text);
+  });
```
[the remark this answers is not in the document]
domw30: Done.

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,112 @@
...
+
+    test('should append spinnerPosition type to text container className attribute', () => {
+      const { container } = renderComponent();
+      const textContainer = container.getElementsByClassName('zui-loading-indicator-text');
+
+      expect(textContainer[0]).toHaveClass(`zui-loading-indicator-text-${MOCK_DEFAULT_PROPS.spinnerPosition}`);
+    });
+  });
+
+  describe('when props are undefined', () => {
```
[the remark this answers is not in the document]
domw30: Done

domw30:
@colbr, ready for another review round - wasn't too sure what was meant here:
_One more test to add - spinnerPosition relies on the spinner being the first element in the container. It uses flex reverse properties to set positioning, so if the spinner isn't first the spinnerPosition will be reversed. Perhaps we could make this easier to test if the markup changed for different spinnerPositions, rather than a class name._

Also, I've re-grouped the tests, but a bit confused where to determine if a test falls into 'class names' or 'spinner position' as they cross over, i.e. spinner position tests focus on changes in class name too.

domw30:
Noticed in the <Card /> testing you've used ```toBeInTheDocument```. Should we use that over ```toBeTruthy``` as I've used in these tests?

colbr:
> Noticed in the testing you've used `toBeInTheDocument`. Should we use that over `toBeTruthy` as I've used in these tests?

I like that `toBeInTheDocument` is a little more explicit than `toBeTruthy`, so I think we should go with that.

colbr:
> wasn't too sure what was meant here: _One more test to add - spinnerPosition relies on...._

I probably didn't write this out clear enough, my bad!

The spinner is positioned by applying different CSS classes. If you look at the top and bottom positioning classes, they use `flex-direction: column` and `flex-direction: column-reverse` respectively.

The ordering of the elements is `<Spinner><div>`. `flex-direction: column` will stack `<Spinner>` on top of `<div>`, and vice versa for `flex-direction: column-reverse`. If you change the ordering of the elements, e.g. `<div><Spinner>`, but not the CSS, `flex-direction: column` (e.g. `spinnerPosition=top`) will stack the `<div>` on top of the `<Spinner>`, which is the opposite of what it should be.

I think it would provide a little bit of extra confidence if we could confirm that the markup is correct for the styling - i.e. that `<Spinner>` comes before `<div>`.

> Perhaps we could make this easier to test if the markup changed for different spinnerPositions, rather than a class name.

This was definitely not very clear, apologies.

I think if, instead of applying a class name for `left`, `right`, `top`, and `bottom`, we could just have a class name for `vertical` and `horizontal`, and instead change the mark up.

So: `spinnerPosition=top`
```
<div class='vertical'>
    <Spinner>
    <div>
</div>
```


So: `spinnerPosition=bottom`
```
<div class='vertical'>
    <div>
    <Spinner>
</div>
```

This would probably be marginally easier to test but probably not worth the added complexity!

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
+
+    expect(loadingIndicatorContainer[0]).toHaveClass(MOCK_DEFAULT_PROPS.className);
+  });
+
+  test('should apply spinnerPosition prop to container class', () => {
+    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
+    const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
+
+    expect(loadingIndicatorContainer[0]).toHaveClass(`zui-loading-indicator-${MOCK_DEFAULT_PROPS.spinnerPosition}`);
+  });
```
colbr, line 73: I like to be explicit with the props relevant to the test, e.g.
```
render(<LoadingIndicator {...DEFAULT_PROPS} spinnerPosition={'top'} />)
```
It's a small readability thing that makes the tests a little bit easier to understand. I'm not too attached to this code style though - if you think it has downsides let me know!

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
+    expect(messageContainer[0]).toHaveClass(`zui-loading-indicator-message-${MOCK_DEFAULT_PROPS.spinnerPosition}`);
+  });
+
+  // undefined
+  test('should not apply className prop to container class', () => {
+    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} className={undefined} />);
+    const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
+
+    expect(loadingIndicatorContainer[0]).not.toHaveClass(MOCK_DEFAULT_PROPS.className);
+  });
```
colbr, line 88: I think this test can be removed, unless I'm missing something?

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -7,29 +7,19 @@ import classNames from 'classnames';
...
+    <div className={classNames('zui-loading-indicator', `zui-loading-indicator-${spinnerPosition}`, className)}>
+      <Spinner />
+      {message && (
         <div
-          className={classNames('zui-loading-indicator-text', `zui-loading-indicator-text-${spinnerPosition}`)}
-          data-testid={TEST_ID.TEXT}
+          className={classNames('zui-loading-indicator-message', `zui-loading-indicator-message-${spinnerPosition}`)}
         >
-          {text}
+          {message}
```
colbr, line 22: We only need to apply the positioning classes to the `zui-loading-indicator` container.

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
+
+    expect(loadingIndicatorContainer[0]).toHaveClass(`zui-loading-indicator-${MockSpinnerPositionType.BOTTOM}`);
+  });
+
+  test('should apply default "bottom" spinnerPosition to message container class', () => {
+    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} spinnerPosition={undefined} />);
+    const messageContainer = container.getElementsByClassName('zui-loading-indicator-message');
+
+    expect(messageContainer[0]).toHaveClass(`zui-loading-indicator-message-${MockSpinnerPositionType.BOTTOM}`);
+  });
```
colbr, line 102: Per previous comment you can drop these tests after you remove the positioning classes from child elements.

colbr (changes requested):
Nice stoked with how this is shaping up. Feels like we're landing on a good testing strategy, thanks for your hard work 🤝

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
+  test('should render spinner component', () => {
+    const { getByText } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
 
-test('should render text', () => {
-  const { getByTestId } = renderComponent({ text: mockText });
-  const text = getByTestId(TEST_ID.TEXT);
-  expect(text).toBeInTheDocument();
-  expect(text).toHaveClass('zui-loading-indicator-text');
-  expect(text).toHaveClass('zui-loading-indicator-text-bottom');
+    expect(getByText(MOCK_SPINNER)).toBeTruthy();
```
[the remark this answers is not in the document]
domw30: Updated

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
-  const { getByTestId } = renderComponent({});
-  const spinner = getByTestId(TEST_ID.SPINNER);
-  expect(spinner).toBeInTheDocument();
-});
+    const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
+    expect(loadingIndicatorContainer.length).toBe(1);
+  });
+
+  test('should render spinner component', () => {
+    const { getByText } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
```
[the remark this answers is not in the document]
domw30: Updated

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
+  test('should render message container and message value', () => {
+    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
+    const textContainer = container.getElementsByClassName('zui-loading-indicator-message');
+
+    expect(textContainer.length).toBe(1);
+    expect(textContainer[0].textContent).toEqual(MOCK_DEFAULT_PROPS.message);
+  });
+
+  // undefined
+  test('should not render message container', () => {
```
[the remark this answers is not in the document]
domw30: Oops sos forgot to remove these! Updated.

src/components/LoadingIndicator/LoadingIndicator.test.tsx
```diff
@@ -1,48 +1,103 @@
...
-  const { getByTestId } = renderComponent({ text: mockText, spinnerPosition: 'left' });
-  const text = getByTestId(TEST_ID.TEXT);
-  expect(text).toHaveClass('zui-loading-indicator-text-left');
+describe('class names', () => {
+  // defined
+  test('should apply className prop to container class', () => {
+    const { container } = render(<LoadingIndicator {...MOCK_DEFAULT_PROPS} />);
+    const loadingIndicatorContainer = container.getElementsByClassName('zui-loading-indicator');
+
+    expect(loadingIndicatorContainer[0]).toHaveClass(MOCK_DEFAULT_PROPS.className);
```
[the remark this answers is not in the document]
domw30: Nice I do think it should be switched yeah. Updated this and other tests!

src/components/LoadingIndicator/LoadingIndicator.test.tsx
domw30: Roger that, I'm all for that. Updated. 

You'll see I've used an enum for spinnerPosition value. Error throws if I use a hardcoded string like 'top'

src/components/LoadingIndicator/LoadingIndicator.tsx
domw30: Ah Yeah you're right, I should have noticed this wouldn't have been changing anything. Updated!

src/components/LoadingIndicator/LoadingIndicator.test.tsx
domw30: Updated!

domw30:
> 

Nice thanks, yeah understand what you mean with testing the expected direction.

I had a think, and had a go at using `horizontal` and `vertical`, but I don't understand how you'd cover all four directions - top, bottom, right, left. How does the following work with the same class 🤔 

```
  &-horizontal {
    flex-direction: row-reverse;
  }
  ```

```
<div class='horizontal'>
    <Spinner>
    <div>
</div>
```

Say the CSS here results in the direction `<div><Spinner />`, how would one use the same `horizontal` class but get the direction to result in `<Spinner /> <div>`? Without using another class, wouldn't this restrict directions?


https://user-images.githubusercontent.com/39112648/192577838-a5467886-455b-49b1-b7ed-398608f9914f.mov

PR #39, opened 2022-09-22
feat: Add ThemeEngine component (portal)

colbr:
Variation of #20.

- Uses `@radix-ui/react-portal` to append a `<style>` node to `<head>`.
- Splits `ThemeEngine` and `StyleEngine` into two parts - the former for selecting a theme, the latter for converting a theme into CSS.
- Better testing.

### DIFF
--- a/src/components/ThemeEngine/StyleEngine.test.tsx
+++ b/src/components/ThemeEngine/StyleEngine.test.tsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import { StyleEngine } from './StyleEngine';
+import { render } from '@testing-library/react';
+
+let mockPortal = jest.fn();
+
+jest.mock('@radix-ui/react-portal', () => ({
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  Root: (props: any) => {
+    mockPortal(props);
+    return <div data-testid="mock-portal">{props.children}</div>;
+  }
+}));
+
+jest.spyOn(global.console, 'warn');
+
+describe('<StyleEngine />', () => {
+  describe('when using Radix Portal (@radix-ui/react-portal)', () => {
+    const { getByTestId } = render(<StyleEngine />);
+
+    test('should render a style element in portal', () => {
+      const head = getByTestId('mock-portal');
+      expect(head.children.length).toBe(1);
+      expect(head.children[0].nodeName).toBe('STYLE');
+    });
+
+    test('should set isChild=true on portal', () => {
+      expect(mockPortal).toHaveBeenCalledWith(expect.objectContaining({ asChild: true }));
+    });
+
+    test('should set container=document.head on portal', () => {
+      expect(mockPortal).toHaveBeenCalledWith(expect.objectContaining({ container: document.head }));
+    });
+  });
+
+  describe('when all CSS is valid', () => {
+    test('should add formatted CSS variables to :root', () => {
+      const styles = { primaryColor1: '#000', alphaPrimary1: 'rgba(177, 78, 255, 0.03)' };
+      const expectedCSS = ':root { --zui-primary-color-1: #000; --zui-alpha-primary-1: rgba(177, 78, 255, 0.03); }';
+
+      const { container } = render(<StyleEngine styles={styles} />);
+      const style = container.querySelector('#zui-theme-variables');
+      expect(style.innerHTML).toBe(expectedCSS);
+    });
+  });
+
+  describe('when passing invalid CSS', () => {
+    const styles = { primary: 'styleengine', secondary: 'red' };
+    const expectedCSS = ':root { --zui-secondary: red; }';
+    let container: HTMLElement;
+    beforeEach(() => {
+      jest.resetAllMocks();
+      const result = render(<StyleEngine styles={styles} />);
+      container = result.container;
+    });
+
+    test('should filter invalid CSS', () => {
+      const style = container.querySelector('#zui-theme-variables');
+      expect(style.innerHTML).toBe(expectedCSS);
+    });
+
+    test('should warn in console', () => {
+      expect(console.warn).toHaveBeenCalledTimes(1);
+    });
+  });
+});
--- a/src/components/ThemeEngine/StyleEngine.tsx
+++ b/src/components/ThemeEngine/StyleEngine.tsx
@@ -1,0 +1,76 @@
+import { CSS_PREFIX } from '../constants';
+import { kebabCase } from 'lodash';
+import React, { ReactNode, useMemo } from 'react';
+
+import { Root as PortalRoot } from '@radix-ui/react-portal';
+
+/**
+ * Validates a CSS color variable.
+ * @param value The value to validate
+ */
+const isColor = (value: string) => {
+  const s = new Option().style;
+  s.background = value;
+  return s.background !== '';
+};
+
+/**
+ * Adds children to document.head via Radix Portal.
+ * @param children The children to add to document.head
+ */
+const HeadPortal = ({ children }: { children: ReactNode }) => {
+  return (
+    <PortalRoot asChild={true} container={document.head}>
+      {children}
+    </PortalRoot>
+  );
+};
+
+type Styles = {
+  [key: string]: string;
+};
+
+/**
+ * Converts an object of key/value pairs into a CSS string.
+ * Values are validated to ensure they are valid CSS color variables.
+ * Variable names are formated to kebab-case and have --zui appended.
+ * @param styles
+ */
+const mapStylesToCSS = (styles: Styles): string => {
+  return Object.entries(styles)
+    .map(([style, value]) => {
+      // Theme values are currently all colors - check they are valid
+      // colors before adding them to the output styles
+      if (isColor(value)) {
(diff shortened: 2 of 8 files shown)


src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { kebabCase } from 'lodash';
+import React, { ReactNode, useMemo } from 'react';
+
+import { Root as PortalRoot } from '@radix-ui/react-portal';
+
+/**
+ * Validates a CSS color variable.
+ * @param value The value to validate
+ */
+const isColor = (value: string) => {
```
joonchoi0918, line 11: It will be better to add return type `boolean`

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const s = new Option().style;
+  s.background = value;
+  return s.background !== '';
```
joonchoi0918, line 14: I am wondering about this:

```
a.style.background = "url(https://web-dev.imgix.net/image/admin/OIF2VcXp8P6O7tQvw53B.jpg?auto=format&w=845)";

=> a.style.background = 'url("https://web-dev.imgix.net/image/admin/OIF2VcXp8P6O7tQvw53B.jpg?auto=format&w=845")'
```

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+/**
+ * Adds children to document.head via Radix Portal.
+ * @param children The children to add to document.head
+ */
+const HeadPortal = ({ children }: { children: ReactNode }) => {
```
joonchoi0918, line 21: Seems like `HeadPortal` is `React.FC`

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <PortalRoot asChild={true} container={document.head}>
```
joonchoi0918, line 23: And I don't think we need to pass `true` value for `boolean` props:

`<PortalRoot asChild={true}>` === `<PortalRoot asChild>`

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+      // Theme values are currently all colors - check they are valid
+      // colors before adding them to the output styles
+      if (isColor(value)) {
+        return `--${CSS_PREFIX}-${kebabCase(style)}: ${value};`;
+      } else {
+        console.warn('CSS variable not supported', style, value);
+        return;
+      }
+    })
+    .join(' ')
```
joonchoi0918, line 51: I think we need to add filter here to remove `undefined` (invalid color) .

```
...
.map(....)
.filter(Boolean)
...
```

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Adds an array of CSS key value pairs to a <style> tag in <head>.
+ * Also appends converts CSS keys to the format --zui-<key-as-kebab-case>
+ * @param styles
+ * @constructor
+ */
+export const StyleEngine = ({ styles }: StyleEngineProps) => {
+  const mappedStyles = useMemo(() => {
+    return styles ? mapStylesToCSS(styles) : '';
```
joonchoi0918, line 68: How about to add styles parameter validation in `mapStylesToCSS(styles)`?

```
const mapStylesToCSS = (styles?: Styles): string => {
   if (!styles) return '';
   ....
}
```

Then we can use 
```
const mappedStyles = useMemo(() => mapStylesToCSS(styles), [styles]);
```

joonchoi0918 (approved):
Nice work!

src/components/ThemeEngine/StyleEngine.tsx
colbr: Themes are currently just color values, so I think we're good to keep this at least for now.

src/components/ThemeEngine/StyleEngine.tsx
colbr: Yeah I probably should use `FC` - I'll make a note of it.

src/components/ThemeEngine/StyleEngine.tsx
colbr: Yeah true - not sure which is preferrable. I have been being explicit with them in the past but I'm opening to dropping `={true}` in situations like this.

src/components/ThemeEngine/StyleEngine.tsx
colbr: Hmm yeah probably right, good spot.

src/components/ThemeEngine/StyleEngine.tsx
colbr: I generally like to avoid adding optional parameters to helper functions because they can make components a little more difficult to understand. I'll have a think in the morning.

colbr:
Note: this one is reasonable experimental, still some polishing to do but the idea is there.

jtulloch (changes requested):

src/components/ThemeEngine/StyleEngine.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      container = result.container;
+    });
+
+    test('should filter invalid CSS', () => {
+      const style = container.querySelector('#zui-theme-variables');
+      expect(style.innerHTML).toBe(expectedCSS);
+    });
+
+    test('should warn in console', () => {
+      expect(console.warn).toHaveBeenCalledTimes(1);
```
jtulloch, line 63: i think this needs to be more explicit, or it doesn't really tell us much. eg. .. if we care that there was a warning, we should be testing what exactly that warning is, so we can be confident that it is actually the warning we think it is.

src/components/ThemeEngine/StyleEngine.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  describe('when passing invalid CSS', () => {
+    const styles = { primary: 'styleengine', secondary: 'red' };
+    const expectedCSS = ':root { --zui-secondary: red; }';
+    let container: HTMLElement;
+    beforeEach(() => {
+      jest.resetAllMocks();
+      const result = render(<StyleEngine styles={styles} />);
+      container = result.container;
```
jtulloch, line 54: i don't think setup should be in `beforeEach` it's much clearer to have this in the test, so you don't have to scroll around trying to figure out what exactly the test is acting on.

jtulloch (changes requested):

src/components/ThemeEngine/StyleEngine.tsx
colbr: Sweet, if y'all both prefer `asChild` I'll change it over.

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+      // colors before adding them to the output styles
+      if (isColor(value)) {
+        return `--${CSS_PREFIX}-${kebabCase(style)}: ${value};`;
+      } else {
+        console.warn('CSS variable not supported', style, value);
+        return;
+      }
+    })
+    .join(' ')
+    .toString()
```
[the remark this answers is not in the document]
colbr: Oops, yeah, you're right - I was using `toString` without `join` initially, and forgot to remove the `toString`. Good spotting.

src/components/ThemeEngine/StyleEngine.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * @constructor
+ */
+export const StyleEngine = ({ styles }: StyleEngineProps) => {
+  const mappedStyles = useMemo(() => {
+    return styles ? mapStylesToCSS(styles) : '';
+  }, [styles]);
+
+  return (
+    <HeadPortal>
+      <style id={'zui-theme-variables'} type={'text/css'}>{`:root { ${mappedStyles} }`}</style>
```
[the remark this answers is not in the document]
colbr: Yeah you're probably right there. I think this was a good option to explore but not being able to restrict scope is a pretty major downside.

src/components/ThemeEngine/StyleEngine.test.tsx
colbr: True - I'll adjust it if we end up revisiting this idea.

src/components/ThemeEngine/StyleEngine.test.tsx
colbr: I was going off [what jest recommends](https://jestjs.io/docs/setup-teardown#scoping) with this one, although perhaps this pattern is slightly less clear when rendering components. 

I think it's marginally clearer in an isolated test if you render the component in each test, but imo the whole file is much more digestible if the exact same render steps aren't repeated throughout. Maybe not as big of an issue in this specific instance, but I could see it being helpful in more complicated components.

jtulloch (changes requested):

colbr:
Closing as we're going with the other solution.

PR #40, opened 2022-09-22
feat: update expose of components and add storybook for missing components

joonchoi0918:


src/components/InfoTooltip/InfoTooltip.tsx
```diff
@@ -1,11 +1,9 @@
 import React, { FC } from 'react';
 import './InfoTooltip.scss';
-import { Tooltip } from '../Tooltip';
+import { Tooltip, TooltipProps } from '../Tooltip';
 import { IconQuestionMark } from '../Icons';
 
-export type InfoTooltipProps = {
-  content: string;
-};
+export type InfoTooltipProps = TooltipProps;
```
domw30, line 6: nit: could we do something like `export type InfoTooltipProps extends TooltipProps {}` instead?

domw30 (approved):
Looks good to me @joonchoi0918! A couple of minor comments. It would be worth resolving conflicts before we ping Brett for a merge. 💯 

Could you change PR name - `expose` to `exposure`

PR #44, opened 2022-09-29
test(wizard): improve testing

domw30:
- improve testing of Wizard and Preset component
- increase testing coverage of Wizard and Presets component


src/components/Wizard/Wizard.tsx
```diff
@@ -1,19 +1,15 @@
...
 
-// Preset Screens
-import { Buttons } from './Presets/Buttons';
-import { Confirmation } from './Presets/Confirmation';
-import { Loading } from './Presets/Loading';
 import './Wizard.scss';
 
 export interface HeaderProps {
-  header?: ReactNode;
+  header?: React.ReactNode;
```
domw30, line 12: Only added this as we import React due to lint - 'React' must be in scope when using JSX

domw30:
I extracted Header from Wizard.tsx, but reverted for now as slightly out of scope for task - do you think it should be placed in its own file?

PR #49, opened 2022-10-04
feat: add buttons variants props in wizard buttons

joonchoi0918:


eric-zero-dev (approved):

domw30 (approved):
Nice 👍

colbr:
As mentioned in Discord I don't think passing props to the buttons is how we want this to function. Passing props down will make it difficult to enforce the constraints of the design system.

Instead, we should handle the button types as variations.

PR #58, opened 2022-10-14
feat(toggle-group): add ToggleGroup

colbr:


domw30 (commented):
Looks awesome. Using attributes like this for a selection strategy makes things so much more explicit than classes, I'd prefer us to push forward with this 👍

PR #62, opened 2022-10-17
refactor: update text stack prop type to take react node

domw30:
- update 'text' prop type to ReactNode. This will prevent error when passing in components such as `ArrowLink` etc


colbr (dismissed):
Looks nice. Would it also be worth making `AsyncText` `text` a `ReactNode` as well?

colbr (approved):

PR #66, opened 2022-10-18
feat: add Image component

colbr:
Adds a component for smooth Image loading. Shows a `Skeleton` until loading is finished, then fades the loaded image in. If image is loading from browser cache, skip the fade in.

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -1,0 +1,63 @@
+import React, { useEffect } from 'react';
+
+import { Skeleton } from '../Skeleton';
+
+import styles from './Image.module.scss';
+import classNames from 'classnames';
+
+export interface ImageProps {
+  alt: string;
+  src?: string;
+  className?: string;
+  objectFit?: 'contain' | 'cover' | 'fill' | 'none' | 'scale-down';
+  onLoad?: () => void;
+  onError?: () => void;
+}
+
+interface ImageState {
+  isCached: boolean;
+  src: string;
+}
+
+// named Img here to avoid conflict with the HTMLImageElement in useEffect
+const Img = ({ className, objectFit = 'cover', alt, src, onLoad, onError }: ImageProps) => {
+  const [image, setImage] = React.useState<ImageState | undefined>();
+
+  useEffect(() => {
+    setImage(undefined);
+    if (src && Boolean(new URL(src))) {
+      const image = new Image();
+      image.src = src;
+      // if image is cached, set isCached so we can skip animations
+      if (image.complete) {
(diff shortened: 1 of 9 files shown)


domw30 (dismissed):
Components and tests look good, no comments from me 💯

padala (dismissed):
Looking good :)

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+interface ImageState {
+  isCached: boolean;
+  src: string;
+}
+
+// named Img here to avoid conflict with the HTMLImageElement in useEffect
+const Img = ({ className, objectFit = 'cover', alt, src, onLoad, onError }: ImageProps) => {
+  const [image, setImage] = React.useState<ImageState | undefined>();
```
joonchoi0918, line 24: nit: no need to specify `undefined` in state type.

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,63 @@
...
+      if (image.complete) {
+        setImage({ isCached: true, src });
+      } else {
+        // else wait until load is complete to set state
+        image.onload = () => {
+          setImage({ isCached: false, src });
+        };
+      }
+    }
+  }, [src]);
```
joonchoi0918, line 41: `src` is props and set it to state again. I think it will be replaced by `isLoaded: boolean`

src/components/Image/Image.tsx
colbr: I didn't know that - thanks!

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <div
+      className={classNames(styles.Container, className)}
+      data-status={image?.src ? 'loaded' : 'loading'}
```
joonchoi0918, line 46: How about to write the state like this?

```
interface ImageState {
  isCached: boolean;
  isLoaded: boolean;
}
.....
const [image, setImage] = React.useState<ImageState>();

....
setImage({ isCashed: false, isLoaded: true});

....
data-status={image?.isLoaded? 'loaded' : 'loading'}
```

joonchoi0918 (dismissed):
Looks Great! 💯 
Just left a few nit comments about my suggestion, but it is up to you.

src/components/Image/Image.tsx
colbr: Yeah I'm not sure which is best to be honest. I like that storing the loaded image in state makes it quite explicit what image you're showing.

I renamed `image` to `loadedImage` for better clarity there!

PR #71, opened 2022-11-15
feat: add select input

ashmidgley:
[Create SelectInput Component for zUI](https://www.notion.so/rational-nomads/Create-SelectInput-Component-for-zUI-ea881e61993d4d49ba4a59c6160e1e25)

-----

## 1. Pull request checklist

## 2. PR type
```
- Feature
```

Missing SelectInput component needed for DAO launcher flow.

Addition of SelectInput component.

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -12,6 +12,8 @@ export interface InputProps extends Omit<AriaTextFieldProps, 'value' | 'onChange
   className?: string;
   wrapperClassName?: string;
   inputClassName?: string;
+  alertClassName?: string;
+  helperTextClassName?: string;
   error?: boolean;
   // @deprecated
   success?: boolean;
@@ -44,6 +46,8 @@ export const Input = forwardRef<HTMLDivElement, InputProps>(
       className,
       wrapperClassName,
       inputClassName,
+      alertClassName,
+      helperTextClassName,
       endEnhancer,
       startEnhancer,
       error,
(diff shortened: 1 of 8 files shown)


src/components/Input/Input.tsx
```diff
@@ -10,6 +10,10 @@ import styles from './Input.module.scss';
 
 export interface InputProps extends Omit<AriaTextFieldProps, 'value' | 'onChange'> {
   className?: string;
+  wrapperClassName?: string;
+  inputClassName?: string;
+  alertClassName?: string;
+  helperTextClassName?: string;
```
colbr, line 16: nit: I generally don't like exposing all the internal class names as it will most likely lead to components being restyled (and thus breaking the design system). It looks like `SelectInput` does a relatively small amount of restyling on the base `Input` component - perhaps we could do this a different way which avoids exposing extra class props.

colbr (changes requested):
Only one required change - nesting the `scss` classes. Looks good all round 🎸

src/components/Input/Input.tsx
ashmidgley: Yeah, I agree. Extra className props in `Input` are a bit messy :(

Is this okay as a temp solution? Once the `DropdownMenu` 2.0 changes go through we can delete this `SelectInput` and remove the additional classNames from `Input`.

ashmidgley:
Cheers for the review @colbr. Have pushed changes to nest the scss classes!

src/components/Input/Input.tsx
colbr: Yep we'll assess it as an iteration

colbr (approved):

PR #74, opened 2022-11-23
fix: remove loading header placeholders

domw30:
[Notion Task](https://www.notion.so/zerotech/4985d8ec0c2f4eb4860678eec483e6f6?v=40a33a299f3541f18ec40fbf914ba46f&p=39a32e9780e448ac9b9a7ddffdb2f976&pm=s)

Task: Static table headers should not show a skeleton when table is loading


colbr (approved):
Beauty

PR #75, opened 2022-11-27
feat(slider): add Slider component

Tweemo:
https://www.notion.so/rational-nomads/Slider-component-005226e174bc4ccaae9507f66d77c74f

-----

## 1. Pull request checklist

## 2. PR type
    - Feature

Previously there was no slider component

Now it has a functioning standard slider and range slider component that offers different sizes


src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,78 @@
...
+.LightRootContainer {
+  color: black;
+}
+
+.SliderRoot {
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: 18.75rem;
+  background-color: #37273f;
```
domw30, line 16: Could we add the colours used to the colors variables file please 🙂 Keep them in line with what they are called in the figma designs

Screenshot 2022-11-30 at 11 26 07

Screenshot 2022-11-30 at 11 28 14

src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,78 @@
+.RootContainer {
```
domw30, line 1: Nice, the styles are look good. A preference I do have is nesting the class names so its easier at first glance to see whats being used and where. I'll add an example below. Don't necessarily have to follow this but if you think its worthwhile feel free to update :)

So based on lines 27 - 47 in `RangeSilder.tsx`, we have a number of elements, all wrapped in div tags with RootContainer as the class name. SliderRoot, SliderTrack etc are nested within. I like to write the style file in similar fashion.

```
.RootContainer {
  //// styles
  .LightRootContainer {
    //// styles
  }

  .SliderRoot {
    //// styles
    .SmallRoot {
      //// styles
    }

    .LightRoot {
      //// styles
    }

    .SliderTrack {
      //// styles
      .SliderRange {
        //// styles
      }
    }
  }
}
```

Something like that. A suggestion but no worries :)

domw30 (changes requested):
Nice work, just a couple of minor requested changes 🙂

src/components/Slider/Slider.module.scss
Tweemo: Done Shifted styles to match their structure in the component

src/components/Slider/Slider.module.scss
Tweemo: I have added them in, but if you could check that the naming convention is what you were expecting that would be awesome.

src/components/Slider/Slider.module.scss
domw30: Sweet its all good. This style helps me quite a lot, was sharing just in case it made sense to you too, not needed at all though only a personal touch

src/components/Slider/Slider.module.scss
```diff
@@ -13,7 +15,7 @@
   display: flex;
   align-items: center;
   width: 18.75rem;
-  background-color: #37273f;
+  background-color: #{$purple-light-4};
```
domw30, line 18: Awesome thanks Tim!

I'm fairly certain you could do `background-color: $purple-light-4` rather than `#{$purple-light-4}`. Not a big deal, as long as either work 🙂

domw30 (dismissed):
Awesome 🚀

src/components/Slider/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,49 @@
...
+const cx = classNames.bind(styles);
+
+export interface RangeSliderProps {
+  values?: number[];
+  min: number;
+  max: number;
+  step: number;
+  minStep?: number;
+  isSmall?: boolean;
+  isLight?: boolean;
```
colbr, line 21: Don't need to worry about theming - apologies, this should have been mentioned in the spec.

src/components/Slider/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,49 @@
...
+      {currentValues[0]} - {currentValues[1]}
+      <form>
+        <SliderRoot
+          className={cx(styles.SliderRoot, { SmallRoot: isSmall, LightRoot: isLight })}
+          defaultValue={[min, max]}
+          min={min}
+          max={max}
+          step={step}
+          minStepsBetweenThumbs={minStep}
+          aria-label="Range slider"
```
colbr, line 37: `aria-label` is used to describe what an element is for - `aria-label` should be provided by the parent to give context, e.g. `aria-label="Temperature slider"`, `aria-label="Age slider"`. If there were 5 range sliders on a page, screen readers might have a tough time figuring out which is which if they're all called `"Range slider"`.

src/components/Slider/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,49 @@
...
+          onValueChange={value => setCurrentValues(value)}
+        >
+          <SliderTrack className={styles.SliderTrack}>
+            <SliderRange className={cx(styles.SliderRange, { SmallRange: isSmall, LightRange: isLight })} />
+          </SliderTrack>
+          <SliderThumb className={cx(styles.SliderThumb, { SmallThumb: isSmall, LightThumb: isLight })} />
+          <SliderThumb className={cx(styles.SliderThumb, { SmallThumb: isSmall, LightThumb: isLight })} />
```
colbr, line 44: `cx(styles.SliderThumb, { SmallThumb: isSmall, LightThumb: isLight })` is repeated - best to set it as a variable in the function body imo.

```ts
const thumbClasses = cx(styles.SliderThumb, { SmallThumb: isSmall, LightThumb: isLight });

return  <SliderThumb className={thumbClasses} />
```

However, I think it's best to handle these cases as custom attributes rather than classes, e.g.
```ts
// component
<SliderThumb data-size={isSmall ? "small" : "default"} />

// selecting in scss
.SliderThumb[data-size="small"] { ... }
```
or similar.

src/components/Slider/Slider.module.scss
colbr: I'd like to get them nested before approving this one - nested styles are much more maintainable. Let me know if you need assistance here @Tweemo.

src/components/Slider/StandardSlider/StandardSlider.tsx
```diff
@@ -0,0 +1,49 @@
...
+  Thumb as SliderThumb
+} from '@radix-ui/react-slider';
+
+export interface SliderProps {
+  step?: number;
+  min?: number;
+  max?: number;
+  value?: number;
+  className?: string;
+  onValueChange?: (value: number) => void;
```
colbr, line 20: Noticed that this prop isn't being used anywhere - it should be passed to `onValueChange` on `SliderRoot`.

colbr (changes requested):
Looking good for the most part - I think the main thing we need to change is collapsing `RangeSlider` and `StandardSlider` into one component. The difference in functionality isn't large enough to warrant splitting into two components.

We need to think about how this component would be used - currently the state is internal, how would you access it from the parent?

Also worth noting that we're using SCSS modules, which means class names can be very simple. For example, `RootContainer` can just be `Container`, `SliderRoot` can just be `Root`, `SliderRange` can just be `Range`, etc. etc.

Left a few scattered comments, but the above 3 are the main things! We'll probably also need to bulk out the tests - I think we're missing some critical functionality, but not much point touching them until the above are fixed.

Nice work!

src/components/Slider/Slider.module.scss
Tweemo: Managed to get them nested and looking good, some handy learning was done :D

src/components/Slider/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
```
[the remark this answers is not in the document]
Tweemo: Merged both components into one slider that takes an `isRangeSlider` prop

src/components/Slider/RangeSlider/RangeSlider.tsx
```diff
@@ -0,0 +1,49 @@
...
+  isSmall?: boolean;
+  isLight?: boolean;
+}
+
+export const RangeSlider: React.FC<RangeSliderProps> = ({ step, min, max, values, isSmall, minStep, isLight }) => {
+  const [currentValues, setCurrentValues] = useState(values ?? [min, max]);
+  return (
+    <div className={cx(styles.RootContainer, { LightRootContainer: isLight })}>
+      {currentValues[0]} - {currentValues[1]}
+      <form>
```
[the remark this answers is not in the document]
Tweemo: The documentation says that its so the events from the input thumbs are propagated properly. https://www.radix-ui.com/docs/primitives/components/slider

src/components/Slider/RangeSlider/RangeSlider.tsx
Tweemo: @colbr As in should I just remove the isLight property and associated code or?

src/components/Slider/RangeSlider/RangeSlider.tsx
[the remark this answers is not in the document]
colbr: I see - makes sense. I think you can remove the `<form>` here, I think the documentation is saying that it will change the thumb elements to `input`s if rendered in a `form`, and outside of a `form` they'll be something else (probably `div`).

src/components/Slider/RangeSlider/RangeSlider.tsx
colbr: Yep you can remove `isLight` and all of the associated code - our theme engine creates CSS variables so no need to handle it in JS, you just use the CSS variables and they'll change based on the theme.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export const Slider: React.FC<SliderProps> = ({
+  step,
+  min,
+  max,
+  value,
+  isSmall,
+  isLight,
+  isRangeSlider,
```
colbr, line 30: Is `isRangeSlider` necessary here? It looks like a range slider requires `min` and `max` props to be defined, so `isRangeSlider` seems functionality equivalent to `has min and max`. I could be missing something here, but looks like we `isRangeSlider === Boolean(min) && Boolean(max)`, in which case we should remove it as it makes the slider a bit more complicated to configure. If there's something I'm missing let me know, there might be some edge cases I'm not considering.

If you wanted to do some Typescript trickery, you could do something like
```ts
interface SliderProps {
    // all values apart from min and max
}

interface RangeSliderProps extends SliderProps {
    min: number;
    max: number;
}

export const Slider = ({ ... }: SliderProps | RangeSliderProps) => ...
```

This would ensure that a standard Slider wouldn't have `min` and `max`, and a range slider will always need both `min` and `max`. I can't remember the exact Typescript syntax so might be slightly off!

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export interface SliderProps {
+  step?: number;
+  min?: number;
+  max?: number;
+  value?: number;
+  className?: string;
+  minStep?: number;
+  isLight?: boolean;
+  isSmall?: boolean;
```
colbr, line 17: Sorry I missed this one on the first review - can we change this to `size: 'default' : 'small'`. The prop should default to `default`, i.e.
```ts
export const Slider = ({
    ...
    size = 'small',
    ...
})
```

This keeps us open to possible design changes in the future, for example if they add a `large` variant.

src/components/Slider/Slider.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  render(<Slider {...DEFAULT_PROPS} isSmall />);
+  expect(mockSliderRange).toHaveBeenCalledWith(expect.objectContaining({ className: 'SliderRange SmallRange' }));
+});
+
+test('should pass props to Radix Slider Thumb', () => {
+  render(<Slider {...DEFAULT_PROPS} />);
+  expect(mockSliderThumb).toHaveBeenCalledWith(expect.objectContaining({ className: 'SliderThumb' }));
+});
+
+test('should correctly render both thumbs', () => {
```
colbr, line 81: `should correctly render both thumbs when isRangeSlider=true`

colbr (changes requested):
Couple small things. Apologies for missing `isSmall` on the first iteration, still a bit dusty from COVID!

Looking good though, excited to get this one merged.

src/components/Slider/Slider.tsx
Tweemo: Both sliders will need a min and max as they dictate the range that the slider functions within. The only difference is with the rangeSlider, it uses the min and max as default values as well.

colbr: I see - makes sense! All good as is then 😊

src/components/Slider/Slider.module.scss
Tweemo: Done

src/components/Slider/RangeSlider/RangeSlider.tsx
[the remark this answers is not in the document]
Tweemo: ohh right, Ill remove it then

src/components/Slider/RangeSlider/RangeSlider.tsx
Tweemo: oh perfect

src/components/Slider/Slider.test.tsx
Tweemo: Done

src/components/Slider/Slider.tsx
Tweemo: @colbr Ive just pushed up my interpretation of this, hope its right, otherwise lemme know :D

colbr (approved):
Nice!

colbr (approved):

PR #77, opened 2022-12-13
feat: add button variant to modal dialog trigger

domw30:
- add `buttonVariant` prop to modal dialog trigger to expose `variant` prop on the button component

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -17,14 +17,15 @@ import classNames from 'classnames';
 export interface ModalProps extends DialogProps {
   className?: string;
   trigger?: ReactNode;
+  buttonVariant?: 'primary' | 'secondary' | 'negative' | 'text';
 }
 
-export const Modal: React.FC<ModalProps> = ({ className, children, trigger, ...rest }) => {
+export const Modal: React.FC<ModalProps> = ({ className, children, trigger, buttonVariant, ...rest }) => {
   return (
     <DialogRoot {...rest}>
       {trigger && (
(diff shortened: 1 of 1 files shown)


colbr (changes requested):
Can we change the interface a little so that the `buttonVariant` prop only exists if `trigger` isn't passed in? Can't remember the exact syntax, but something like

```ts
interface ModalProps {
    // all but trigger and button variant
}

interface ButtonModalProps {
    buttonVariant?: 'primary' | 'secondary' | 'negative' | 'text';
}

interface TriggerModalProps {
    trigger?: ReactNode;
}

export const Modal: React.FC<ButtonModalProps | TriggerModalProps>
```

domw30:
> Can we change the interface a little so that the `buttonVariant` prop only exists if `trigger` isn't passed in? Can't remember the exact syntax, but something like
> 
> ```ts
> interface ModalProps {
>     // all but trigger and button variant
> }
> 
> interface ButtonModalProps {
>     buttonVariant?: 'primary' | 'secondary' | 'negative' | 'text';
> }
> 
> interface TriggerModalProps {
>     trigger?: ReactNode;
> }
> 
> export const Modal: React.FC<ButtonModalProps | TriggerModalProps>
> ```

Hey @colbr, just checked this out as I'll need to access the variant thats used for disabled state of the button, and not sure I understand why we’d want buttonVariant prop only to exist if trigger isn't passed in, slightly confused as for buttonVariant to be exposed it requires trigger. 

Screenshot_2022-12-15_at_09 51 09

PR #72, opened 2022-11-21
feat: standard checkbox

pnagarjun16:
[Standard Checkbox component](https://www.notion.so/rational-nomads/Standard-Checkbox-component-b5a2532bb50f4e02bcfae4b9ce7e8deb)

-----

## 1. Pull request checklist
- [x] Notion card has been moved to the Code Review column
- [x] Notion card has a link to this PR
- [x] A reviewer has been assigned to the Notion card

## 2. PR type

    - Feature

## 3. What is the old behavior?
No common checkbox component avaialble

## 4. What is the new behavior?
Added new checkbox component

PR #80, opened 2023-01-18
feat(status badge): status badge component

pnagarjun16:
[Status badge component](https://www.notion.so/rational-nomads/Status-Badge-component-1937fdf656c94714a40c2f51b1ba9351)

-----

## 1. Pull request checklist

## 2. PR type
    - Feature

The status badge component is not available

Implemented the status badge component


src/components/index.ts
```diff
@@ -23,3 +23,4 @@ export * from './ToggleGroup';
 export * from './Tooltip';
 export { Wizard } from './Wizard';
 export * from './Checkbox';
+export * from './Status';
```
domw30, line 26: nit: could we keep these in alphabetical order for consistency please 🙂

src/styles/_colors.scss
```diff
@@ -7,6 +7,7 @@ $dark-secondary: #05201b;
 $grey: #808080;
 $grey-3: #232326;
 $grey-9: #706f78;
+$grey-10: #7e7d86;
```
domw30, line 10: Nice, thanks for adding these

src/components/Status/Status.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+.Status {
+  display: flex;
+  width: 8px;
+  height: 8px;
+  border-radius: 9999px;
+  background: var(--zui-status-background);
+}
```
domw30, line 31: I like the structure of this file, but I think we could simplify and make it a little more readable. If your up for it could you try something like the following and let me know what you think?

I've changed the order slightly and removed the var(). Also I don't think the `#{}` is required for the colours...I could be wrong though. Could you also try replacing `border-radius: 9999px;` with `50%` for the same outcome?

```

.Status {
	display: flex;
	width: 8px;
	height: 8px;
	border-radius: 50%;

	&[data-variant='active'] {
		background: $green-9;
	}

	&[data-variant='idle'] {
		background: $orange-10;
	}

	&[data-variant='busy'] {
		background: $read-10;
	}

	&[data-variant='offline'] {
		background: $grey-10;
	}

	&[data-variant='unread'] {
		background: $purple-11;
	}
}

```

domw30 (changes requested):
Awesome work @pnagarjun16 ! I made a few suggestions. Let me know what you think/any questions 🚀

src/components/index.ts
pnagarjun16: Updated

src/components/Status/Status.module.scss
pnagarjun16: Changed the border radius to %. Changed size from px to rems. As per brett's comments If there is a variation of styles for the same element we are declaring the variables accordingly ([link](https://github.com/zer0-os/zUI/pull/72#discussion_r1038658400) ).  If we are using imported colors inside the variable we use `#{}`.

domw30: Looks good. I'll aim to look through this again over the weekend. Thanks for making those other changes !

domw30: Looking good to me, just one small request - rather than declaring `.Status {}` twice could we nest all styles into one? Very minor, but doing so will improve readability, I haven't run this myself but this should have the same outcome 

```
.Status { 
   display: flex; 
   width: 8px; 
   height: 8px; 
   border-radius: 50%
   
   &[data-variant='active'] {
     background: $green-9; 
    } 
    
    &[data-variant='idle'] {
      background: $orange-10; 
     }
     
     ....
```

src/components/Status/Status.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { useState } from 'react';
```
domw30, line 1: Looks like `useState` can be removed from this file

domw30 (changes requested):
A couple of minor changes and we're good to go! Thanks @pnagarjun16

src/components/Status/Status.tsx
pnagarjun16: Removed

domw30 (approved):
Nice work @pnagarjun16 !!

PR #82, opened 2023-02-05
feat(avatar): avatar component

pnagarjun16:
[Avatar Component](https://www.notion.so/rational-nomads/Avatar-component-5cbedb3121a44409b50a3e3a7e7c32c9)

-----

## 1. Pull request checklist

## 2. PR type
    - Feature

The Avatar component is not available

Integrated the new avatar component

### DIFF
--- a/src/components/Avatar/Avatar.module.scss
+++ b/src/components/Avatar/Avatar.module.scss
@@ -1,0 +1,180 @@
+@import '../../styles/colors';
+
+.Avatar {
+  &:focus-within {
+    --zui-avatar-border-radius: 50%;
+  }
+
+  .AvatarRoot {
+    --zui-avatar-border-radius: 50%;
+    --zui-avatar-width: 2.5rem;
+    --zui-avatar-height: 2.5rem;
+  }
+
+  &[data-type='square'] {
+    .AvatarRoot {
+      --zui-avatar-border-radius: 0.5rem;
+    }
+
+    &:focus-within {
+      --zui-avatar-border-radius: 0.5rem;
+    }
+  }
+
+  .AvatarFallback {
+    --zui-avatar-font-size: 0.875rem;
+    --zui-avatar-font-line-height: 150%;
+  }
+
+  .Badge {
+    --zui-avatar-badge-position-right: -0.375rem;
+    --zui-avatar-badge-position-top: -0.375rem;
+  }
+
+  &[data-size='small'] {
+    .AvatarRoot {
+      --zui-avatar-width: 1.5rem;
+      --zui-avatar-height: 1.5rem;
+    }
+
+    .Badge {
+      --zui-avatar-badge-position-top: -0.625rem;
+      --zui-avatar-badge-position-right: -0.625rem;
+    }
+
+    .AvatarFallback {
+      --zui-avatar-font-size: 0.625rem;
+      --zui-avatar-font-line-height: 0.75rem;
+    }
+  }
+
+  &[data-size='extra small'] {
+    .AvatarRoot {
+      --zui-avatar-width: 1rem;
+      --zui-avatar-height: 1rem;
+    }
+
+    .AvatarFallback {
+      --zui-avatar-font-size: 0.5rem;
+      --zui-avatar-font-line-height: 0.625rem;
+    }
+  }
+
+  &[data-size='medium'] {
+    .AvatarRoot {
+      --zui-avatar-width: 2rem;
+      --zui-avatar-height: 2rem;
+    }
+
+    .Badge {
+      --zui-avatar-badge-position-top: -0.625rem;
+      --zui-avatar-badge-position-right: -0.625rem;
+    }
+
+    .AvatarFallback {
+      --zui-avatar-font-size: 0.625rem;
+      --zui-avatar-font-line-height: 0.75rem;
+    }
+  }
+
+  .Status {
+    &[data-status-type='active'] {
+      --zui-avatar-status-background: #{$green-9};
+    }
+
+    &[data-status-type='idle'] {
+      --zui-avatar-status-background: #{$orange-10};
+    }
+
+    &[data-status-type='busy'] {
+      --zui-avatar-status-background: #{$read-10};
+    }
+
+    &[data-status-type='offline'] {
+      --zui-avatar-status-background: #{$grey-10};
+    }
+
+    &[data-status-type='unread'] {
+      --zui-avatar-status-background: #{$purple-11};
+    }
+  }
+}
+
+.Avatar {
+  position: relative;
+  display: flex;
+  cursor: pointer;
+
+  &:focus-within {
+    border-radius: var(--zui-avatar-border-radius);
+    outline-offset: 0.25rem;
+    outline: 0.0625rem solid $green-11;
+  }
+
+  .AvatarRoot {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    vertical-align: middle;
+    overflow: hidden;
+    user-select: none;
+    background-color: $black;
+    border-radius: var(--zui-avatar-border-radius);
+    width: var(--zui-avatar-width);
+    height: var(--zui-avatar-height);
+  }
+
+  .AvatarImage {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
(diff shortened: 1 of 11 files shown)


src/components/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,180 @@
...
+  display: flex;
+  cursor: pointer;
+
+  &:focus-within {
+    border-radius: var(--zui-avatar-border-radius);
+    outline-offset: 0.25rem;
+    outline: 0.0625rem solid $green-11;
+  }
+
+  .AvatarRoot {
```
colbr, line 114: You can remove `Avatar` from these classes - the module will make sure they're scoped properly.

Just `.Avatar {  .Root { } }` is fine!

src/components/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  &[data-type='square'] {
+    .AvatarRoot {
+      --zui-avatar-border-radius: 0.5rem;
+    }
+
+    &:focus-within {
+      --zui-avatar-border-radius: 0.5rem;
+    }
```
colbr, line 21: You should be able to declare these CSS variables outside of the classes, so just in the top-level class.

```
.Avatar {
    &[data-type='square'] {
        --zui-avatar-border-radius: 0.5rem;
    }
}
```

Same goes for most of the others, except for the status ones! Unless you can think of a benefit to keeping them scoped? I may be missing something.

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  // Used to fetch the first two letter of the name
+  useEffect(() => {
+    if (userFriendlyName) {
+      const userNameArray = userFriendlyName.split(' ');
+      let charFromName = userNameArray[0].substring(0, 1);
+      charFromName += userNameArray.length >= 2 ? userNameArray[1].substring(0, 1) : userNameArray[0].substring(1, 2);
+      setAvatarName(charFromName);
+    }
+  }, [userFriendlyName]);
```
colbr, line 36: Move this functionality out to a utility function in `Avatar.tsx`, i.e. `getInitials`, then call the function in the component body? You won't need to save state, just something like `const initials = userFriendlyName  && getInitials(userFriendlyName)`.

I understand using `useEffect` so the computation doesn't run on every render, but `useEffect` comes with its own overheads. If the function you're running as a side-effect isn't very resource intensive (like this one), it's best to just calculate it on render as it makes the component easier to maintain. In fact, it's generally advised to avoid using `useEffect` unless you're syncing with something outside of React - I think a `useMemo` would be more appropriate here if you were needing to save resources.

[There's a little write up here](https://beta.reactjs.org/learn/you-might-not-need-an-effect) that covers this more in-depth!

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,71 @@
...
+interface StatusTypeProps {
+  statusType: AvatarProps['statusType'];
+}
+const AvatarStatus = ({ statusType }: StatusTypeProps) => {
+  return <div className={styles.Status} data-status-type={statusType}></div>;
+};
+
+interface StatusBadgeTypeProps {
+  badgeContent: AvatarProps['badgeContent'];
+}
```
colbr, line 67: Can we use the `Status` component here?

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,71 @@
...
+      let charFromName = userNameArray[0].substring(0, 1);
+      charFromName += userNameArray.length >= 2 ? userNameArray[1].substring(0, 1) : userNameArray[0].substring(1, 2);
+      setAvatarName(charFromName);
+    }
+  }, [userFriendlyName]);
+
+  return (
+    <div className={styles.Avatar} data-type={type} data-size={size} tabIndex={0}>
+      <RadixAvatar.Root className={styles.AvatarRoot}>
+        <RadixAvatar.Image className={styles.AvatarImage} src={imageURL} alt={userFriendlyName} />
```
colbr, line 41: Add `"avatar"` to `alt={userFriendlyName}`, just so it's obvious to screen readers what the image is.

src/components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  size: 'regular',
+  type: 'circle'
+};
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+describe('<Avatar />', () => {
+  test('should have only avatar with regular size and circle shape', () => {
```
colbr, line 16: Can you change the names of these to include some context, i.e.

```
describe 'when using default props'
    test 'should be regular size'
    test 'should be circle shape'

describe 'when status type is "active"'
    test 'should show active badge'
```

colbr (changes requested):
A couple changes, but all-in-all looks great! Nice work, seems like you're getting the hang of zUI.

src/components/Avatar/Avatar.module.scss
pnagarjun16: Updated

src/components/Avatar/Avatar.module.scss
pnagarjun16: I updated the fallback-related styles to the root level. But the other styles are not effecting if we move it to top-level class.

src/components/Avatar/Avatar.tsx
pnagarjun16: Thanks for the info. I updated it to `const initials = userFriendlyName  && getInitials(userFriendlyName)`

src/components/Avatar/Avatar.tsx
pnagarjun16: Removed these and added the Status component

src/components/Avatar/Avatar.tsx
pnagarjun16: Updated

src/components/Avatar/Avatar.test.tsx
pnagarjun16: updated the test cases

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+interface StatusBadgeTypeProps {
+  badgeContent: AvatarProps['badgeContent'];
+}
+
+const AvatarBadge = ({ badgeContent }: StatusBadgeTypeProps) => {
+  return <div className={styles.Badge}> {badgeContent}</div>;
+};
+
```
colbr, line 56: nit: Add a doc comment here for `getInitials` function

src/components/Avatar/Avatar.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    expect(container.querySelector('[data-status-type="active"]')).toBeInTheDocument();
+  });
+});
+
+describe('when badge content is "9+"', () => {
+  test('should show badge', () => {
+    const { getByText } = render(<Avatar {...DEFAULT_PROPS} badgeContent="9+" />);
+    expect(getByText('9+')).toBeInTheDocument();
+  });
+});
```
colbr, line 49: Oh, I missed a test case in my last review. Can we also write a test to ensure a user's initials are visible if there's no badge content. Can we also write tests to make sure different name lengths work, i.e. `John` appears as `J`, `John Smith` appears as `JS`, and `John Smith Appleseed` appears as `JA`?

colbr (changes requested):
Missed a test case in my last review. Add that test then we can move it through!

src/components/Avatar/Avatar.tsx
pnagarjun16: Updated

src/components/Avatar/Avatar.test.tsx
pnagarjun16: Added the missed test cases. We are always displaying two letters from the user name. As per figma if there are three words we are fetching the first letters from the first two words.
Screenshot 2023-02-13 at 21 39 16

colbr: Awesome, good stuff!

colbr (approved):

PR #83, opened 2023-02-10
feat: Add IconButton component

dalefukami:
## 1. Pull request checklist

- [x] N/A Jira task has been moved to the Code Review column
- [x] N/A Link to Jira task at the top of this PR
- [x] N/A Jira task has a link to this PR

## 2. PR type
 

Feature

Adds an IconButton component for when an icon is meant to be actionable.


jest.setup.ts
```diff
@@ -1 +1,6 @@
 import '@testing-library/jest-dom/extend-expect';
+import 'jest-extended/all';
+import { configure } from 'enzyme';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
+
+configure({ adapter: new Adapter() });
```
colbr, line 6: Can we write tests using `react-testing-library`, e.g. [Card tests](https://github.com/zer0-os/zUI/blob/master/src/components/Card/Card.test.tsx), [Button tests](https://github.com/zer0-os/zUI/tree/master/src/components/Button)? If not we should agree on a standard so we're not doing half-enzyme-half-RTL.

src/components/IconButton/IconButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  cursor: pointer;
+  background: none;
+  border: none;
+  display: inline-block;
+  margin: 0px;
+
+  &:hover {
+    background-color: theme.$color-greyscale-transparency-3;
+  }
+}
```
colbr, line 17: I didn't do a design review in Storybook but it looks like the `:active` styles from Figma are missing here.

[Icon Button in Figma](https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/ZERO-Design-System-(zDS)?node-id=5043%3A106692&t=zdrU2dk3zXSbz2z0-1)

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { JSXElementConstructor } from 'react';
+
+import styles from './IconButton.module.scss';
+import { IconProps } from '../Icons/Icons.types';
+
+export interface Properties {
+  className?: string;
+  onClick: () => void;
+
```
colbr, line 9: Needs a minimal variant ([Figma](https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/ZERO-Design-System-(zDS)?node-id=5043%3A106692&t=zdrU2dk3zXSbz2z0-1))

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  Icon: JSXElementConstructor<IconProps>;
+  label?: string;
+  size?: string | number;
+  isFilled?: boolean;
+}
+
+export class IconButton extends React.Component<Properties> {
```
colbr, line 16: Can we write this as a function component to be consistent with the other components in zUI?

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styles from './IconButton.module.scss';
+import { IconProps } from '../Icons/Icons.types';
+
+export interface Properties {
+  className?: string;
+  onClick: () => void;
+
+  Icon: JSXElementConstructor<IconProps>;
+  label?: string;
+  size?: string | number;
```
colbr, line 12: What are the valid `string`s for `size` here?

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export class IconButton extends React.Component<Properties> {
+  handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
+    event.stopPropagation();
+    this.props.onClick();
+  };
+
+  render() {
+    return (
+      <button className={`${styles.IconButton} ${this.props.className}`} onClick={this.handleClick}>
```
colbr, line 24: `classNames(styles.IconButton, this.props.className)`, because when concatenating strings if the `className` prop is undefined the resulting class name will be  `IconButton_something undefined`.

colbr, line 24: Wondering if we could reuse the zUI `Button` component in some way here, as it definitely has some shared functionality with `IconButton`.

src/components/IconButton/IconButton.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  line-height: 0px;
+
+  cursor: pointer;
+  background: none;
+  border: none;
+  display: inline-block;
+  margin: 0px;
+
+  &:hover {
+    background-color: theme.$color-greyscale-transparency-3;
```
colbr, line 15: Is there a benefit here to using SCSS `use` vs. using the CSS variable directly (`--color-greyscale-transparency-3)?

colbr (changes requested):
Some scattered questions here. Happy to discuss any of them and work out how we both want to write components.

I also added a PR template if you'd like to copy + paste it! It will be automatically applied to new PRs, but for this one you'll have to drag it over 😊

src/components/IconButton/IconButton.module.scss
dalefukami: Oh, strange, those are different than the one [I was sent](https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/ZERO-Design-System-(zDS)?node-id=8718%3A43245&t=iqkZtphiZs1SMdsr-0).

I'll confirm.

src/components/IconButton/IconButton.tsx
dalefukami: Yeah, this is one that I couldn't decide on. Feels heavy handed to use a library just to make that check when having `undefined` in the class list isn't really a big deal (other than it looks ugly under the hood). I actually wonder what impact that has on rendering performance :thinking: That'd be interesting to know.

I'll move back to `classNames`. Though, I wonder if there's a better way for us to make this base case simpler given that we do it so often on the root node of a component. My first thought when I see it is usually, "Oh, what are the conditional classes we get on this?" and, of course, in this case there are none. So, it's wasted scanning time.

src/components/IconButton/IconButton.module.scss
dalefukami: The reason I went that way as that the build will fail if you use a value that doesn't exist. If you use the css var directly then you really won't know. It's still possible the theme will reference a css var that doesn't exist but it felt cleaner to have that part all collected in a single place. We might even be able to auto-generate the theme file?

jest.setup.ts
dalefukami: I can't seem to figure out how to do shallow rendering in RTL. Maybe you can show me an example? The Button tests seem to mock the sub component but still perform a full render on the mock.

src/components/IconButton/IconButton.tsx
dalefukami: Same as an [Icon](https://github.com/zer0-os/zUI/blob/master/src/components/Icons/Icons.types.ts#L1).

src/components/IconButton/IconButton.tsx
dalefukami: :thinking: I think the duplication is fairly limited. Maybe if the two evolve in similar ways often enough then we can identify what's worth sharing but it seems premature to jump straight to that to me.

src/components/IconButton/IconButton.tsx
dalefukami: Sure. This component is simple enough.

dalefukami:
> I also added a PR template...

Updated the description but I think we might want to alter the checklist:
1. Jira tasks are highly custom to teams so states may not be the same
2. This assumes one PR per Jira task which is an unnecessary restriction
3. For an "open source" repo, having links to private Jira tasks is somewhat strange. Probably better to require thorough descriptions? I'm not really sure why our repos are public but, since they are, it probably makes sense to not link to private things.

src/components/IconButton/IconButton.module.scss
colbr: Ah yes I see - the one you were sent is the `Icon` component, which is different to `IconButton` by the looks of it. `IconButton` is under `IconGroup`, so a little bit hard to find!

jest.setup.ts
colbr: We aren't shallow rendering in our tests - I generally prefer to test things as close as possible to how a user will use them. I'm definitely open to the idea of using shallow rendering though, as long as we're careful about which implementation details we're testing (if any).

I think I chatted to Joel about this a while ago, maybe we need to start the conversation again! If you're happy to use RTL for now we can push on and chat async, otherwise we can expedite the conversation.

src/components/IconButton/IconButton.module.scss
colbr: Cool - nice idea that seems very helpful! We'll look into theme engine next sprint

colbr:
> > I also added a PR template...
> 
> Updated the description but I think we might want to alter the checklist:
> 
>     1. Jira tasks are highly custom to teams so states may not be the same
> 
>     2. This assumes one PR per Jira task which is an unnecessary restriction
> 
>     3. For an "open source" repo, having links to private Jira tasks is somewhat strange. Probably better to require thorough descriptions? I'm not really sure why our repos are public but, since they are, it probably makes sense to not link to private things.

1. True - I was absent for a little while so only just got onboarded to our Jira yesterday. The main reason I like having ticket links in the PR is so if I get a review notification in Github, I can easily find which Jira ticket it's related to. However, I imagine this can be solved better than linking to the ticket.
2. Same story as above, this is more relic of how we did things in Notion. I'll chat with the team re: how to improve this!
3. Yes I am a little confused about that myself - probably another thing to discuss.

colbr (approved):

src/components/IconButton/IconButton.module.scss
dalefukami: Yeah, turns out the design group is gonna tweak the designs I was given and I'll follow up with a new PR probably. I'm going to close this one until then.

src/components/IconButton/IconButton.tsx
dalefukami: Yeah, it's reasonably minimal. I try to tend towards just not relying on external libs in the cases when it's not necessary (not just for the performance impact).

In any case, I do think if we're going to have this be a common thing across components, adding a class and className prop to the root node every time, then it might be worth seeing if there's a clean way to get that mostly for free every time we create a new component. I'm just not sure if there's a clean way to do it that wouldn't just make things more confusing. We'd have to experiment a bit, I think.

PR #84, opened 2023-02-10
chore: Update the classnames module and fix errors

dalefukami:
## 1. Pull request checklist

- [x] N/A Jira task has been moved to the Code Review column
- [x] N/A Link to Jira task at the top of this PR
- [x] N/A Jira task has a link to this PR

## 2. PR type
 

Other: Update library version.

## 5. Other information

This updates the `classnames` library to the next patch version.

The library changed their typescript definition of the `classnames/bind` import to explicitly require an object to bind to as the first parameter. We were importing the `bind` version in some cases where we weren't actually using it for that purpose. I updated the imports as necessary and removed the ones where the library itself wasn't actually required.


colbr (approved):
Good spot!

PR #85, opened 2023-02-13
fix: resolve missing key console errors

domw30:
[Associated Jira Task 1](https://wilderworld.atlassian.net/jira/software/c/projects/ZAP/boards/34?modal=detail&selectedIssue=ZAP-44)
[Associated Jira Task 2](https://wilderworld.atlassian.net/jira/software/c/projects/ZAP/boards/34?modal=detail&selectedIssue=ZAP-48)

-----

## 1. Pull request checklist

- [ ] Jira task has been moved to the Code Review column
(Unable to action the above due to restriction on Jira for subtask)
- [x] Link to Jira task at the top of this PR
- [x] Jira task has a link to this PR

## 2. PR type
- Bugfix

AsyncTable row and grid placeholders are missing keys which is resulting in console errors being thrown across the zApps using the component.

Screenshot 2023-02-13 at 13 34 38

Screenshot 2023-02-14 at 16 45 03

The addition of a key on row/grid placeholder components prevents related console errors being thrown.

### DIFF
--- a/src/components/AsyncTable/Placeholder.tsx
+++ b/src/components/AsyncTable/Placeholder.tsx
@@ -1,8 +1,6 @@
 import React from 'react';
 import Skeleton from 'react-loading-skeleton';
 
-import { randomUUID } from '../../utils';
-
 import { TableData } from './Column';
 
 interface RowPlaceholdersProps {
@@ -18,11 +16,10 @@ export const RowPlaceholders = ({ amount, height, numColumns }: RowPlaceholdersP
     </TableData>
   );
 
-  const rowPlaceholderKey = `grid-placeholder-key-${randomUUID()}`;
   const columns = Array(numColumns).fill(columnPlaceholder);
   const rows = Array(amount).fill(<tr>{columns}</tr>);
 
-  return <div key={rowPlaceholderKey}>{rows}</div>;
+  return <div>{rows}</div>;
 };
 
 interface GridPlaceholdersProps {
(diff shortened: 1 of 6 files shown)


domw30:
@colbr - attempt to resolve these console errors with a random string as key. Let me know if this could be improved. Using the randomUUID util will mean the keys change - do you see that as problematic?

src/components/AsyncTable/Placeholder.tsx
```diff
@@ -1,31 +1,38 @@
...
-      <Skeleton width={'100%'} height={height} />
-    </TableData>
-  );
-
-  const columns = Array(numColumns).fill(columnPlaceholder);
-  const rows = Array(amount).fill(<tr>{columns}</tr>);
+  const columns = Array(numColumns)
+    .fill(null)
+    .map((_, index) => (
+      <TableData key={index} alignment={'center'}>
```
domw30, line 16: The reason for removing `columnPlaceholder` here was to prevent the error:

`<tr> cannot appear as a child of <div>`

I would have needed to wrap `div` tags around `columnPlaceholders` to add the key attribute and index, which would result in the above error.

src/components/AsyncTable/RowPlaceholders.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default {
+  title: 'Placeholder Elements/RowPlaceholders',
+  component: RowPlaceholders
+} as ComponentMeta<typeof RowPlaceholders>;
+
+const Template: ComponentStory<typeof RowPlaceholders> = args => {
+  return (
+    <StoryCard isContrast>
+      <RowPlaceholders {...args} />
```
domw30, line 15: The following error is displayed when running in storybook, however, after some investigation this error is not thrown when running a zApp using the Table component. I believe this may be something to do with Storybook rendering the components wrapped in div tags? Either way this shouldn't occur when using these components in zApps.

Screenshot 2023-02-20 at 15 53 43

PR #88, opened 2023-02-22
feat: Add Theme Component

dalefukami:
## 1. Pull request checklist

## 2. PR type
 

Feature

None

Added a ThemeComponent and updated storybook to allow toggling between the two existing themes.


src/components/IconButton/IconButton.tsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import styles from './IconButton.module.scss';
 import { IconProps } from '../Icons/Icons.types';
 
-export interface Properties {
+export interface IconButtonProperties {
```
dalefukami, line 7: Made this change because I discovered the generic `Properties` can't be used since we export these all as one huge index file and I got name collisions :). I can pull this into a separate PR if you'd prefer.

dalefukami:
There seems to be some lint failures on the `main` branch that are causing my PR to fail. I'll fix those up in a separate PR.

src/components/ThemeEngine/ThemeEngine.tsx
```diff
@@ -0,0 +1,43 @@
...
+  Dark = 'dark'
+}
+
+export interface ThemeEngineProperties {
+  viewMode: ViewModes;
+  element: HTMLElement;
+  theme: { [viewMode: string]: { [styleProp: string]: string } };
+}
+
+export class Component extends React.Component<ThemeEngineProperties> {
```
colbr, line 17: nit: could simplify this down to a function component, something like below

```ts
import React from 'react';
import kebabCase from 'lodash.kebabcase';

import theme from './theme.json';

export enum ViewModes {
  Light = 'light',
  Dark = 'dark'
}

type Theme = { [styleProp: string]: string };
type Themes = { [viewMode: string]: Theme };

export interface ThemeEngineProperties {
  viewMode: ViewModes;
  element: HTMLElement;
  theme: Themes;
}

export const ThemeEngine = React.memo(
  ({ theme = ViewModes.Dark, element = document.documentElement, themes = theme }: ThemeEngineProperties) => {
    writeThemePropertiesToElement(themes[theme], element);
    return null;
  }
);

const writeThemePropertiesToElement = (theme: Theme, element: HTMLElement) => {
  Object.keys(themeProperties).forEach(cssProperty => {
    element.style.setProperty(`--${kebabCase(cssProperty)}`, themeProperties[cssProperty]);
  });
};
```

colbr (changes requested):
Looking good nice work!

The required `viewMode` is the only requested change, everything else is a suggestion/nit 🚀

src/components/ThemeEngine/ThemeEngine.tsx
```diff
@@ -0,0 +1,43 @@
...
+      this.props.element.style.setProperty(`--${kebabCase(prop)}`, modeObject[prop]);
+    });
+  }
+
+  render() {
+    return <></>;
+  }
+}
+
+export function ThemeEngine(props: Partial<ThemeEngineProperties>) {
```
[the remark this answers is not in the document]
dalefukami: I don't think we can prevent anyone from setting the mode to undefined explicitly. But, I made it required and prevented the export of the underlying component as we're not really meant to allow people to modify anything but which theme is selected. I think that resolves the issues. The viewMode (now 'theme') is already an enum so it's not possible to set "red". I'm not sure how you were able to do that.

colbr (approved):

PR #89, opened 2023-02-23
fix: Fix CSS lint errors

dalefukami:
## 1. Pull request checklist

## 2. PR type
 

Bugfix

We're receiving various lint errors such as:
`79:1  ✖  Unexpected duplicate selector ".Accordion",   no-duplicate-selectors`

No more lint errors

## 5. Other information

I verified by running storybook locally and comparing the UI behaviour/look against the production storybook. Everything appears to match.

### DIFF
--- a/src/components/Accordion/Accordion.module.scss
+++ b/src/components/Accordion/Accordion.module.scss
@@ -1,80 +1,6 @@
 @import '../../styles/colors';
 
-.Accordion {
-  --zui-accordion-margin-top: 0.25rem;
-  --zui-accordion-group-border-radius: 0.5rem;
-
-  .Item:first-child {
-    --zui-accordion-margin-top: 0;
-  }
-
-  .Header {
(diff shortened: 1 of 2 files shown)


src/components/Accordion/Accordion.module.scss
```diff
@@ -109,36 +36,87 @@
   .Header {
     display: flex;
     width: 100%;
-    border-bottom: var(--zui-accordion-border-bottom);
+    border-bottom: 0.063rem solid #{$purple-50};
 
     &:hover {
       cursor: pointer;
+      border-bottom: 0.063rem solid #{$purple-50};
```
domw30, line 43: nit: `#{}` shouldn't be required - `$purple-50`. Looks like these were just moved around though rather than new additions so not to worry.

domw30 (approved):
Ran locally and ran the lint/tests - everything looks good. Thanks for picking that up Dale 💪

PR #93, opened 2023-03-08
add modal confirmation

mrval5:
## 1. Pull request checklist

- add modal confirmation

modal confirmation storybook: https://deploy-preview-93--zui-storybook.netlify.app/?path=/story/data-display-modal--confirmation

## 5. Other information

Figma: https://www.figma.com/file/aETpyuG2AKjNcQtCjldcX0/zOS-Messenger-%26-Notifications?node-id=6391-243627&t=CN0jibtqxVH0OTp0-0


src/components/Modal/ModalConfirmation.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onConfirm,
+  ...rest
+}) => {
+  return (
+    <Modal className={classNames(styles.WiderContent, className)} {...rest}>
+      <div className={styles.Head} data-testid="modal-head">
+        <DialogTitle className={styles.Title}>{title}</DialogTitle>
+
+        <DialogClose asChild className={styles.CloseIcon}>
+          <IconButton Icon={IconXClose} onClick={onCancel} />
```
mrval5, line 36: Hi @colbr , I tried to reuse the icon that we have in the storybook with white color, but I'm not sure why it's now black instead of why. do you know what is the issue ?

dalefukami: Took a quick look. I think it's the default browser style on Button components. We probably should have a global reset style or something so that buttons inherit the color.

dalefukami: image

mrval5: I forgot to add which button I wanted to reuse from storybook: https://zui-storybook.netlify.app/?path=/story/inputs-iconbutton--close-icon.

mrval5: > image

you are right, I've removed the color.

src/components/Modal/Modal.module.scss
```diff
@@ -22,4 +23,36 @@
   border: 1px solid #{$grey-dark};
 
   background: #{$almost-black};
+
+  position: relative;
+}
+
+.WiderContent {
```
domw30, line 30: Hey @mrval5 - we can nest these styles to reflect the component tree like so :

```
@import '../../styles/typography';

.Overlay {
  display: grid;
  position: fixed;
  top: 0;
  left: 0;
  right: 0;
  bottom: 0;

  place-items: center;
  overflow-y: auto;
  z-index: 10000;

  background: rgba(0, 0, 0, 0.9);

  .Content {
    min-width: 18.75rem; // 300px
    padding: 1.875rem; // 30px
    border-radius: 1rem;
    border: 1px solid $grey-dark;

    background: $almost-black;

    position: relative;
  }

  .WiderContent {
    min-width: 37.64rem;

    .Head {
      display: flex;
      justify-content: space-between;
      align-items: center;
      border-bottom: 1px solid $grey-dark;

      margin-bottom: 2.476rem;

      .Title {
        font-size: 1.48rem;
        font-weight: $font-weight-bold;
      }

      .CloseIcon {
        color: inherit;
      }
    }

    .FooterActions {
      display: flex;
      justify-content: space-between;
      align-items: center;

      margin-top: 2.476rem;
    }
  }
}
```

domw30: We try to follow this structure with styling - also placing classes in order they are in the tree, personally I feel this improves the readability. I've ordered the classes and nested here.

src/components/Modal/Modal.module.scss
```diff
@@ -22,4 +23,36 @@
...
+  margin-top: 2.476rem;
+}
+
+.CloseIcon {
+  color: inherit;
+}
+
+.Title {
+  font-size: 1.48rem;
+  font-weight: #{$font-weight-bold};
```
domw30, line 48: nit: We can remove `#{}`

I've done this in the snippet I posted above.

domw30 (changes requested):
Looks good @mrval5 ! I'm going to request a couple of changes, but let me know if you have thoughts on these.

I think it would be a good idea to create a subfolder for `ModalConfirmation` within `Modal` folder. If we add more sub-components of `Modal` this will keep the directory for the component tidy. I'd also suggest adding a `ModalConfirmation.module.scss` file and extract over the styles to that. Sometimes styles get difficult to track using the same parent file.

src/components/Modal/Modal.module.scss
mrval5: Thanks for your review, I've done the changes as suggested

src/components/index.ts
```diff
@@ -16,6 +16,7 @@ export * from './Markdown';
 export * from './MediaInput';
 export * from './Member';
 export { Modal } from './Modal';
+export { ModalConfirmation } from './Modal/ModalConfirmation';
```
domw30, line 19: Maybe we should export `*` here so props can be accessed if needed.

domw30 (changes requested):
Hey @mrval5 thanks for the changes. It looks like there are duplicate `ModalConfirmation` files. Any `ModalConfirmation` files will need to be removed from the `Modal` folder.

Thanks!

domw30 (approved):
Looks good, thanks for the changes!

@mrval5 this should unblock you, apologies for the delay.

PR #95, opened 2023-03-13
fix(grid-card): flicker on image load

colbr:
## 1. Pull request checklist

`GridCard` images would flicker while loading

https://user-images.githubusercontent.com/12437916/224830400-ac4f9518-9acf-42aa-b0e0-0801ea019dc6.mov

`GridCard` was showing a `Skeleton` when `imageSrc` was `undefined`. It would render a `Skeleton`, then would flick to `Image`, which also renders a `Skeleton` until `imageSrc` has loaded. This means it would render one `Skeleton`, then change to another `Skeleton`, before ultimately showing the image.

`GridCard` images don't flicker while loading.

I basically just removed the `Skeleton`, and let `Image` handle `undefined` since it was built for that purpose. 

https://user-images.githubusercontent.com/12437916/224830628-987a58cb-0641-438b-afb2-9186e54d7f95.mov

1. Link to NFTs
2. Navigate the subdomain table on grid view

## 5. Other information

I also did a tiny amount of refactoring while I was at it
- Changed a class name to be a little more descriptive
- Changed `GridCard` to use `forwardRef`

### DIFF
--- a/src/components/GridCard/GridCard.module.scss
+++ b/src/components/GridCard/GridCard.module.scss
@@ -11,7 +11,7 @@
         }
     }
 
-    .ImageContainer {
+    .Aspect {
         display: flex;
         justify-content: center;
         align-items: center;
@@ -26,7 +26,6 @@
             object-fit: cover;
 
             transition: transform 0.3s ease;
-
         }
     }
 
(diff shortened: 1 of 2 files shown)


domw30 (approved):
🔥

PR #97, opened 2023-03-14
feat: implement theming for all components

colbr:
ZAP-73

- Changed existing components to use theme `SCSS` values
- A little bit of refactoring as some components needed more than just a few colours switched
- Changed some tests

**Implications of merging**
zApps may look a bit funky on light mode (no funkier than they already do) until we switch zApps to use theme values.


src/components/Accordion/Accordion.module.scss
```diff
@@ -80,65 +80,71 @@ $group-border-radius: 0.5rem;
...
-
-      &[data-state='open'] {
-        background-color: #{$purple-2};
-      }
-
-      &:hover {
-        background-color: #{$purple-2};
-      }
+    .Content::before {
+      border: 1px solid $color-primary-6;
```
domw30, line 109: should we replace `px` with `0.063rem`

src/components/Accordion/Accordion.module.scss
```diff
@@ -80,65 +80,71 @@ $group-border-radius: 0.5rem;
...
     font-size: 1rem;
     line-height: 1.5rem;
 
+    transition: height 300ms ease;
+    z-index: 0;
+
+    &::before {
+      content: ' ';
+      position: absolute;
+      top: -1px;
```
domw30, line 130: some more uses of `px`, not sure if it's intentional though!

src/components/AsyncTable/RowPlaceholders.stories.tsx
```diff
@@ -1,23 +1,21 @@
...
 const Template: ComponentStory<typeof RowPlaceholders> = args => {
   return (
-    <StoryCard isContrast>
-      <table>
-        <tbody>
-          <RowPlaceholders {...args} />
-        </tbody>
-      </table>
-    </StoryCard>
+    <table>
```
domw30, line 14: Nice I can see why this was previously throwing console errors for the table tags being children of the <div> tag, well... looks like it was the order in `preview.js`.

src/components/DropdownMenu/DropdownMenu.scss
```diff
@@ -12,27 +12,27 @@
...
   padding: 0.5rem;
 
-  background: #{$purple-10};
+  background: $color-primary-2;
   display: flex;
   flex-direction: column;
   border-radius: 1rem;
-  border: 1px solid #{$purple-25};
+
+  // @note: replace hardcoding when we sort shadow variables
```
domw30, line 20: 👍

src/components/InfoTooltip/InfoTooltip.scss
```diff
@@ -13,7 +14,8 @@
     justify-content: center;
     width: 1rem;
     height: 1rem;
-    background-color: #{$grey};
+    background-color: #{$color-greyscale-8};
```
domw30, line 17: nit: should be able to remove `#{}`

src/components/InfoTooltip/InfoTooltip.scss
```diff
@@ -13,7 +14,8 @@
...
+    color: $color-greyscale-12;
```
domw30, line 18: nit: Not a change you've made but could refactor the use of `px` in this file to `rem`

src/components/Markdown/MarkdownEditor/MarkdownEditor.tsx
```diff
@@ -2,16 +2,11 @@ import type { FC } from 'react';
...
-import {
-  MarkdownEditorVariants,
-  MarkdownEditorModes,
-  MARKDOWN_EDITOR_TOOLBAR_TITLES
-} from './MarkdownEditor.constants';
+import { MarkdownEditorModes, MARKDOWN_EDITOR_TOOLBAR_TITLES } from './MarkdownEditor.constants';
 import styles from './MarkdownEditor.module.scss';
-
-const cx = classNames.bind(styles);
+import { Alert } from '../../Alert';
```
domw30, line 9: nit: order of imports could be updated to be in line with zApps - although not sure we follow the structure here in zUI

src/components/Markdown/MarkdownEditor/MarkdownEditor.tsx
```diff
@@ -50,22 +44,23 @@ export const MarkdownEditor: FC<MarkdownEditorProps> = ({
...
+          onChange={onChange}
+          commandsFilter={onCommandsFilter}
+          preview={MarkdownEditorModes.EDIT}
+          textareaProps={{
+            placeholder
+          }}
+        />
+      </div>
+      {errorText && (
+        <Alert variant={'error'} className={styles.ErrorMessage}>
```
domw30, line 60: I haven't looked at the designs for this MDE - is `error` the only variant this could alert could have? i.e. is there a scenario where we'd show a `warning`?

src/components/SelectInput/SelectInput.module.scss
```diff
@@ -1,15 +1,16 @@
 @import '../../styles/colors';
+@import '../../styles/theme';
 
 .InputContainer {
     min-width: 17rem;
     text-align: left;
 
     .Wrapper:hover {
-        border-color: #8E4EC6;
+        //border-color: $color-;
```
domw30, line 9: nit: is this hover style supposed to be removed?

src/components/Slider/Slider.module.scss
```diff
@@ -1,63 +1,72 @@
...
+      height: 1.5rem;
+      width: 1.5rem;
+      background-color: $color-greyscale-12;
+      border: 0.125rem solid $color-secondary-9;
+      border-radius: 1rem;
+      cursor: pointer;
+      outline: none;
+    }
+
+    .Thumb:focus::after {
```
domw30, line 51: nit: we could probs nest this in `Thumb` above. Same with the `Thumb[data-size]` and `Root[data-size]`

src/components/StepBar/StepBar.module.scss
```diff
@@ -1,48 +1,51 @@
...
+    position: absolute;
+    top: calc(100% - 1px);
 
-    &.Hide {
-      border-bottom: 1px solid #{$black};
-      color: #{$black};
-      z-index: -1;
-    }
+    border-bottom: 1px solid $color-primary-9;
+    transition: all 0.2s ease-in-out;
```
domw30, line 49: nit: I think there may be a `0.2s` in `_animations.scss` we could import here.

src/components/StepBar/StepBar.tsx
```diff
@@ -18,44 +14,29 @@ export type StepBarProps = {
...
+        >
+          {s.title}
+        </div>
+      ))}
 
       <div
         style={{
-          width: `${stepStyle.width}%`,
-          transform: `translateX(${stepStyle.translate}%)`
+          width: `${100 / steps.length}%`,
```
domw30, line 35: nit: maybe out of scope for this task but we should add these to the style file in `.Bar {}`

domw30 (approved):
Nice refactoring! Looking really good. Lots of changes and I noticed theres a few `px`'s remaining but all appears to be straight forward enough! Nothing here to prevent merging from what I can see 🚀

src/components/Accordion/Accordion.module.scss
colbr: Good question - imo it makes sense to keep borders `1px` so they don't scale up, but perhaps this is incorrect. I have pinged design team!

src/components/InfoTooltip/InfoTooltip.scss
colbr: Switched `margin-left` to `rem`, I'll check with design about `border-radius`.

src/components/Markdown/MarkdownEditor/MarkdownEditor.tsx
colbr: Good point!

src/components/Markdown/MarkdownEditor/MarkdownEditor.tsx
colbr: Not sure, this is just how it was written initially. Will add a warning state if we come across it

src/components/SelectInput/SelectInput.module.scss
colbr: Good point, I don't think so. Let me check!

colbr: Looks like there's no hover, just `active` and `focus`.

src/components/StepBar/StepBar.module.scss
colbr: May leave it as is now until we sort animation tokens

src/components/StepBar/StepBar.tsx
colbr: Yeah tough to say, because the values rely on JS variables. We'd have to assign CSS variables in the `style` prop to use in CSS. I'll have a think about this one!

domw30 (approved):
Nice one for the responses @colbr. LGTM! 💯

PR #98, opened 2023-03-17
lint: fix warnings

colbr:
per title


domw30 (approved):
Nice, ran lint as well and all looks good.

PR #99, opened 2023-03-21
ModalConfirmation props extends

mrval5:
## 1. Pull request checklist

- [ ] PR name follows [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/)

extends props from the `DialogProps`

### DIFF
--- a/src/components/Modal/ModalConfirmation/ModalConfirmation.tsx
+++ b/src/components/Modal/ModalConfirmation/ModalConfirmation.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Title as DialogTitle, Close as DialogClose } from '@radix-ui/react-dialog';
+import { DialogProps, Title as DialogTitle, Close as DialogClose } from '@radix-ui/react-dialog';
 import classNames from 'classnames';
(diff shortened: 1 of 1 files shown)


src/components/Modal/ModalConfirmation/ModalConfirmation.tsx
```diff
@@ -8,7 +8,7 @@ import { IconXClose } from '../../Icons';
 
 import styles from './ModalConfirmation.module.scss';
 
-export interface ModalConfirmationProps {
+export interface ModalConfirmationProps extends DialogProps {
```
domw30, line 11: Was this preventing modal closing `onCancel` (click cross icon) ?

domw30 (approved):
👍

src/components/Modal/ModalConfirmation/ModalConfirmation.tsx
mrval5: no, it's working good.

I want to use the `open`  prop from `DialogProps` which it was not in `ModalConfirmationProps`

PR #100, opened 2023-03-30
Icon button latest design

mrval5:
## 1. Pull request checklist

- [ ] PR name follows [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/)

I've implemented the new styling of the icon buttons (without tooltip on this PR, it follow later)
Figma: https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/ZERO-Design-System-(zDS)?node-id=3722-12534&t=nJi7LZf1Ywxmm7pb-0
Video explaining the changes: https://www.loom.com/share/713dddcd3acd4a15a38c0e16aba0a7af
Ethan explaining the iconButton https://www.loom.com/share/c8d423a5b4d447f18f5f0247f0732d5a


src/components/IconButton/IconButton.tsx
```diff
@@ -1,28 +1,66 @@
...
+import './IconButton.scss';
+
 export interface IconButtonProperties {
   className?: string;
   onClick: () => void;
 
   Icon: JSXElementConstructor<IconProps>;
   label?: string;
-  size?: string | number;
+  size?: string | number | 'large' | 'small' | 'x-small';
```
domw30, line 14: Nice, we'll have to update where this is used in zApps so we can remove `string|number`. One for later!

domw30 (commented):
Looks good to me @mrval5. Will be good to get @colbr to have a quick look over the `scss`

src/components/IconButton/IconButton.scss
```diff
@@ -0,0 +1,163 @@
...
+  margin: 0px;
+
+  color: theme.$color-greyscale-12;
+
+  &:hover {
+    background-color: theme.$color-greyscale-transparency-3;
+  }
+
+  // Variants
+  &-variant-primary {
```
colbr, line 21: nit: it can be handy to use mix-ins or variables for variants, just so you can define the values for each variant once, then apply the properties once. Not necessary for this PR though!

src/components/IconButton/IconButton.tsx
```diff
@@ -1,28 +1,66 @@
...
+          return 40;
+        case 'small':
+          return 32;
+        case 'x-small':
+          return 24;
+      }
+    }
+
+    return size;
+  };
```
colbr, line 50: nit: move this function outside of the component, e.g.

```ts
const IconButton = () => { ... };

const getSize = (size: IconButtonProperies['size']) => {
    if (size && ['large', 'small', 'x-small'].includes(size.toString())) {
      switch (size) {
        case 'large':
          return 40;
        case 'small':
          return 32;
        case 'x-small':
          return 24;
      }
    }

    return size;
  };
```

colbr (approved):
Nice! Got some questions about modules but I'll dm you, nothing to stop this being approved

src/components/IconButton/IconButton.scss
ratik21: will keep in mind, thanks

src/components/IconButton/IconButton.tsx
ratik21: maybe we can create a common `enum` as well (just to avoid repeating using raw string) ?

src/components/IconButton/IconButton.tsx
ratik21: done.

colbr (approved):

PR #106, opened 2023-04-26
fix: size the alert with border-box

dalefukami:
## 1. Pull request checklist

Due to the `width: 100%` and the padding alerts would overflow the box they were contained in:

With `border-box` `100%` includes all the extra stuff so it gets sized correctly to the container:


colbr (approved):
Good shout - we should really apply `border-box` to every component

dalefukami:
Yeah, I agree. I wasn't quite sure if I should go through them. This one just happened to impact me recently.

PR #115, opened 2023-05-22
feat: input tweaks

colbr:
## 1. Pull request checklist

- Clicking the show/hide button in the password input would show/hide the text, but the button icon wouldn't change
- On Chrome, autofilling an input would apply a white background

- The button icon changes as per Figma
- Autofilled inputs retain normal styles


domw30 (approved):
👍

domw30 (approved):

PR #116, opened 2023-05-22
feat: add toast notification component

domw30:
## 1. Pull request checklist

- toast notification component with animation.

https://github.com/zer0-os/zUI/assets/39112648/eff8a73f-fb7c-4222-ad15-dcd86fcc67c5

- use the zUI deployment preview
- I have used the modal component in the toast story to highlight the intended behaviour of the toast (as an example)

### DIFF
--- a/src/components/ToastNotification/ToastNotification.module.scss
+++ b/src/components/ToastNotification/ToastNotification.module.scss
@@ -1,0 +1,227 @@
+@import '../../styles/theme';
+
+/* reset */
+button {
+  all: unset;
+}
+
+.ToastRoot {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  border-radius: 0.5rem;
+  padding: 1.5rem 0.5rem 1.5rem 1.5rem;
+  gap: 1.5rem;
+
+  .TextWrapper {
+    display: flex;
+    flex-direction: column;
+    max-width: 14.5rem;
+
+    .ToastTitle {
+      font-weight: 700;
+      font-size: 1rem;
+      line-height: 1.5rem;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+
+    .ToastDescription {
+      font-size: 0.75rem;
+      line-height: 0.908rem;
+      color: $color-greyscale-12;
+    }
+  }
+
+  .ToastClose {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 2.25rem;
+    width: 2.25rem;
+
+    &:hover {
+      cursor: pointer;
+    }
+  }
+
+  // primary variants
+  &[data-variant='primary'] {
+    background: $color-primary-2;
+    border: 1px solid $color-primary-6;
+
+    .ToastIcon {
+      color: $color-primary-11;
+    }
+
+    .ToastTitle {
+      color: $color-greyscale-12;
+    }
+
+    .ToastClose {
+      border-left: 1px solid $color-primary-6;
+    }
+  }
+
+  // success variants
+  &[data-variant='success'] {
+    background: $color-success-3;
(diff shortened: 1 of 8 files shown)


src/components/ToastNotification/ToastNotification.module.scss
```diff
@@ -0,0 +1,227 @@
...
+    height: 2.25rem;
+    width: 2.25rem;
+
+    &:hover {
+      cursor: pointer;
+    }
+  }
+
+  // primary variants
+  &[data-variant='primary'] {
```
colbr, line 50: What I like to do for variants is set CSS variables, then you apply the CSS variables once. So like

```scss
&[data-variant="primary"] {
    --background: ...;
    --border-color: ...;
    --icon-color: ...;
    --toast-title: ...;
    --toast-close-border-color: ...;
}

&[data-variant="secondary"] {
    --background: ...;
    --border-color: ...;
    --icon-color: ...;
    --toast-title: ...;
    --toast-close-border-color: ...;
}

background: var(--background);
border: 1px solid var(--border-color);

.ToastIcon {
  color: var(--icon-color);
}

.ToastTitle {
  color: var(--toast-title);
}

.ToastClose {
  border-left: 1px solid var(--toast-close-border-color);
}
```

Just a nit though - not necessary for approval.

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </Toast.Provider>
+  );
+};
+
+ToastNotification.defaultProps = {
+  positionVariant: 'right',
+  themeVariant: 'primary',
+  actionAltText: 'call to action',
+  openToast: false
+};
```
colbr, line 99: I generally like to define these in the prop destructuring, e.g.

```tsx
export const ToastNotification: FC<ToastNotificationProps> = ({
  title,
  description,
  actionTitle,
  actionAltText = 'call to action',
  positionVariant = 'right',
  themeVariant = 'primary',
  swipeDirection,
  openToast = false,
  onClick,
  onClose
}) => {
```

Not necessary for approval though.

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -0,0 +1,99 @@
...
+  actionAltText: string;
+  positionVariant?: ToastPositionVariant;
+  themeVariant?: ToastThemeVariant;
+  swipeDirection?: Toast.SwipeDirection;
+  openToast?: boolean;
+  onClick?: () => void;
+  onClose?: () => void;
+};
+
+export const ToastNotification: FC<ToastNotificationProps> = ({
```
colbr, line 29: Remove `FC` here as it will add `children` as an implicit prop. 

```tsx
export const ToastNotification = ({ ... }: ToastNotificationProps) => ({ ... });
```

src/components/ToastNotification/ToastNotification.module.scss
```diff
@@ -0,0 +1,227 @@
...
+}
+
+@keyframes swipeOutLeft {
+  from {
+    transform: translateX(var(--radix-toast-swipe-end-x));
+  }
+  to {
+    transform: translateX(calc(-100% - var(--viewport-padding)));
+  }
+}
```
colbr, line 227: nit: Wonder if these definitions could be shortened with a SASS function, similar to how we do breakpoints in zApps

colbr (changes requested):
Only required change is removing `FC`, all else looks great nice work!

src/components/ToastNotification/ToastNotification.module.scss
domw30: Ah yeah I think they definitely could be. I'll note to look back on this as a SASS function might be good so we can re-use if needed

PR #118, opened 2023-05-25
fix: `PasswordInput` should not fire `onSubmit` of containing `form` 

colbr:
## 1. Pull request checklist

When using `PasswordInput` inside the `zOS` login form, the show/hide button would be pressed on enter, rather than the `form` submit button. This is because a `button` `type` was not specified for `IconButton`. This also means that any `form` with an `IconButton` inside it wouldn't submit on enter.

- Added a `button` `type` property to `IconButton`
- Set `type` to `type={'button'}` by default
- Added tests to make sure the `type` stays `button`

- Add `PasswordInput` to a `form`
- Add a submit `button` to the `form`
- Type a password input, and hit enter

At this point, the `form` `onSubmit` should have been called, rather than the show/hide button `onClick`.

## 5. Other information

Did a tiny amount of refactoring while I had these files open - you'll see it outlined in the commit history.

### DIFF
--- a/src/components/IconButton/IconButton.test.tsx
+++ b/src/components/IconButton/IconButton.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { render, fireEvent } from '@testing-library/react';
 
 import { IconButton, IconButtonProperties } from '.';
-import { IconXClose } from '../Icons/icons/IconXClose';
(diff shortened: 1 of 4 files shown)


domw30 (approved):
Looks good to me! :rocket:

PR #122, opened 2023-05-26
fix: Remove global reset of buttons

dalefukami:
## 1. Pull request checklist

Removes a global style override of all buttons that changed how buttons were sized. Note that default buttons are rendered with box-sizing->border-box. In general, I don't think we want global overrides happening within specific component stylesheets. With bug:

image

Returned to previous rendering:
image


domw30 (approved):
Good spot! 🚀

PR #123, opened 2023-05-30
feat(button): add enhancers

colbr:
## 1. Pull request checklist

There was no way to add a start or end enhancer to `Button`.

Can now add a start or end enhancer to `Button` through the `startEnhancer` and `endEnhancer` props.

- Check the `Button` enhancer stories in Storybook
- Run tests for `Button.tsx`

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -18,6 +18,9 @@ export interface ButtonProps {
   isLoading?: boolean;
   isDisabled?: boolean;
   isSubmit?: boolean;
+
+  startEnhancer?: ReactNode;
+  endEnhancer?: ReactNode;
 }
 
 export const Button: FC<ButtonProps> = ({
@@ -27,6 +30,8 @@ export const Button: FC<ButtonProps> = ({
   isDisabled,
   isSubmit,
   variant = 'primary',
+  startEnhancer,
+  endEnhancer,
   ...rest
 }) => {
   const ref = useRef(null);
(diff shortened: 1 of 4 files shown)


src/components/Button/Button.tsx
```diff
@@ -52,7 +57,11 @@ export const Button: FC<ButtonProps> = ({
       ...buttonProps
     },
     <>
-      <div className="zui-button-content">{isLoading ? <Spinner className="zui-button-spinner" /> : children}</div>
+      <div className="zui-button-content">
+        {startEnhancer}
+        {isLoading ? <Spinner className="zui-button-spinner" /> : children}
```
dalefukami, line 62: Should the loading state still show the icon? I haven't seen the design but that feels weird to me.

dalefukami (approved):
Just as a general question, do we even need this. Why shouldn't the consumer be in control of what's rendered inside the button by just passing in whatever nodes they want as children?

colbr:
> Just as a general question, do we even need this. Why shouldn't the consumer be in control of what's rendered inside the button by just passing in whatever nodes they want as children?

Oops sorry missed this question. It's standard to pass enhancers in as props - it makes for a simpler API, and encourages correct use of the design system (e.g. the design system specifies 1 enhancer maximum on each side, which would be much less obvious if we dropped `startEnhancer`/`endEnhancer` and left it up to the consumer)

colbr:
Merging now as I have Dale's review. We should change branch protection rules ☺️

PR #127, opened 2023-06-14
feat(search-input): add component

colbr:
## 1. Pull request checklist

`search` `Input` didn't match Figma.

`search` `Input` now matches Figma.

1. Open Storybook
2. Navigate to `Inputs > Search`
3. Check that the component matches [Figma](https://www.figma.com/file/wTOVNZonl9MmEzORwBHDuq/ZERO-Design-System-(zDS)?type=design&node-id=4546%3A36921&t=TObYKB5R5LIbt8S5-1)

### DIFF
--- a/src/components/Input/SearchInput.module.scss
+++ b/src/components/Input/SearchInput.module.scss
@@ -1,0 +1,28 @@
+@import '../../styles/theme';
+
+@mixin path($color, $styles) {
+    svg {
+        path {
+            stroke: $color;
+        }
+    }
+}
+
+.SearchWrapper {
+    border-radius: 1000px !important; // make it pill shaped
+}
+
+.SearchIcon {
+    &[data-is-searching] {
+        svg path {
+            stroke: $color-greyscale-12;
+        }
+    }
+
+    svg path {
+        stroke: $color-greyscale-11;
+        transition: stroke 0.1s ease-in-out;
+    }
+
+    padding-left: 8px;
+}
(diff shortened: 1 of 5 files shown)


domw30 (approved):
Nice 👍

src/components/Input/SearchInput.module.scss
```diff
@@ -0,0 +1,28 @@
...
+@mixin path($color, $styles) {
+    svg {
+        path {
+            stroke: $color;
+        }
+    }
+}
+
+.SearchWrapper {
+    border-radius: 1000px !important; // make it pill shaped
```
ratik21, line 12: what does `!important` do?

src/components/Input/SearchInput.module.scss
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.SearchIcon {
+    &[data-is-searching] {
```
ratik21, line 16: so here `data-is-searching` is a boolean value, and the css is picked on the basis of it being `true`/`false`?

ratik21 (approved):
had 1-2 css doubts

src/components/Input/SearchInput.module.scss
colbr: It overrides previous set values for `border-radius` - for some reason this `border-radius` is being overridden. Will have to work out why it's getting overridden in another PR, but it's not critical right now.

src/components/Input/SearchInput.module.scss
colbr: Yup! It's basically saying `if data-is-searching has a value`, and in the component we set `data-is-searching={isSearching ? '' : null}` ☺️

ratik21: got it, thanks

PR #131, opened 2023-06-22
feat: implement 'isActive' state in avatar component

domw30:
The reason for this change is so we can use Avatar as the trigger for the DropdownMenu component with an `isActive` style. The `focus-within` styling isn't enough to currently satisfy this, as shown in the recordings below.

The issue we're currently facing (recording below) is because the focus-within pseudo class only applies to an element that has received focus itself, or has a descendant which has received focus. However, when the DropdownMenu opens, the Avatar loses its focus because focus is now given to the dropdown menu.

The solution is to manage a state to track whether the dropdown menu is opened or not, and then conditionally add a class to the Avatar to simulate :focus-within.

Here I have created a state variable called isActive in the Avatar component. This means in the DropdownMenu we can set its value with the help of the onOpenChange prop, and use this state variable to conditionally add the focused class to the Avatar component.

## 1. Pull request checklist

Unable to set an active state on Avatar.

Able to set an active state when using Avatar.
Also a small change, removing outline from dropdown trigger.

Avatar used with DropdownMenu before change looks like this:

https://github.com/zer0-os/zUI/assets/39112648/76f5bc79-732a-4e67-b017-dd4c860bd30b

What the Avatar used with DropdownMenu after change will look like:

https://github.com/zer0-os/zUI/assets/39112648/ea8cb7a3-803c-4d2e-ad08-1a06806457e0

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -33,7 +33,7 @@ export const Avatar = ({
 
   return (
     <div
-      className={classNames(styles.Avatar, { [styles.Active]: isActive })}
+      className={classNames(styles.Avatar, { [styles.isActive]: isActive })}
       data-type={type}
       data-size={size}
       tabIndex={0}
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,11 +1,14 @@
 import React from 'react';
 
 import * as RadixAvatar from '@radix-ui/react-avatar';
+
+import { Status } from '../Status';
 import { IconCurrencyEthereum } from '../Icons';
 
 import { AVATAR_ICON_SIZE } from './AvatarIconSize.constants';
+
+import classNames from 'classnames';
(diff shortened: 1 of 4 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
 
 import * as RadixAvatar from '@radix-ui/react-avatar';
+
+import { Status } from '../Status';
```
ratik21, line 5: why do we leave spacings here?

ratik21 (dismissed):

src/components/Avatar/Avatar.tsx
domw30: To group the imports :) In zUI we had a pattern to follow. This is just to keep the imports clean and easily readable. I think we should do it in zOS, but obviously not really a priority!

ratik21: ok makes sense :+1:

PR #133, opened 2023-07-03
feat: add radix dropdown menu arrow pointer

domw30:
The arrow pointer will only be visible if `showPointer` is true, and `alignMenu` is set to center. The reason for this is because the pointer will not line up with the `DropdownMenu` if the alignment is `start` or `end` due to the border radius we have on the `DropdownMenu`. For this case, design would have to provide further input for a solution on how to line up the pointer. As this is not required for our needs right now, I am adding what is necessary for the current requirements.

https://github.com/zer0-os/zUI/assets/39112648/de1415a8-1335-4718-b1c4-bed62fba0c15

Alignment issue that would need to be addressed if we ever require this (using red to highlight):

Screenshot 2023-07-03 at 17 55 22

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.stories.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.stories.tsx
@@ -11,7 +11,7 @@ export default {
 
 const Template: ComponentStory<typeof DropdownMenu> = args => {
   return (
-    <div style={{ margin: '300px' }}>
+    <div style={{ display: 'flex', justifyContent: 'center', alignItems: 'center', height: '100vh', width: '100%' }}>
       <DropdownMenu
         {...args}
         items={[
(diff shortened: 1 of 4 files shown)


src/components/DropdownMenu/DropdownMenu.stories.tsx
```diff
@@ -11,26 +11,28 @@ export default {
...
-          onSelect: () => console.log('you clicked Orange!')
-        },
-        {
-          id: 'pear',
-          label: 'Pear',
-          onSelect: () => console.log('you clicked Pear!')
-        }
-      ]}
-    />
+    <div style={{ display: 'flex', justifyContent: 'center', alignItems: 'center', height: '100vh', width: '100%' }}>
```
domw30, line 14: Wrapped in div to center the elements on the storybook page so we are able to see all alignments and side settings.

dalefukami (dismissed):
"Pointer" made me think of the the cursor but... I dunno if there's a better name. lol

colbr (approved):

PR #140, opened 2023-08-11
fix: icon button sizing and icon sizing to match

domw30:
When passing `size` to the `IconButton`, the button itself is not sized correctly around the `Icon`. This PR fixes this issue by adding an inline style to the button wrapped around the `Icon` component which uses the `size` prop to determine its `width` and `height`.

Now, in zOS (or zApps), you can pass in a size `large`, `small`, `x-small` or simply a number of pixels i.e. something larger like - `48`, or something smaller like - `12`. 

Example - if `large` is set as the `size` for `IconButton`, the `Icon` svg itself will be `40` in width and height, and the button wrapped around the `Icon` will also have a width of `40` and height of `40`.

Examples:
CUSTOM SMALL (`12` number passed in)
Screenshot 2023-08-11 at 12 37 18

CUSTOM LARGE (`68` number passed in)
Screenshot 2023-08-11 at 12 37 13

`x-small` passed in
Screenshot 2023-08-11 at 12 37 09

`small` passed in
Screenshot 2023-08-11 at 12 37 05

`large` passed in
Screenshot 2023-08-11 at 12 36 58

### DIFF
--- a/src/components/IconButton/IconButton.scss
+++ b/src/components/IconButton/IconButton.scss
@@ -1,14 +1,16 @@
 @use '../../styles/theme' as theme;
 
 .zui-iconButton {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
   border-radius: 50%;
   padding: 2px;
   line-height: 0px;
-
   cursor: pointer;
   background: none;
   border: none;
-  display: inline-block;
   margin: 0px;
 
   color: theme.$color-greyscale-12;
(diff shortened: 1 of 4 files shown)


domw30:
@colbr - tagging you to make you aware of this change -  in the `zApp`'s, anywhere we are using `IconButton` this _shouldn't_ break the UI. Let me know if you spot anything off.

colbr (approved):

PR #178, opened 2024-03-04
Remove slide up animation from `Accordion`

ratik21:
**NOTE**: This is a temporary change

Currently we're having issue when using accordion in zOS in the backup-faq modal. You can open a question, but when you "close" it, then the slideUp animation is not working. 

This PR temporarily removes it, until we get the animation working.


domw30 (approved):
👍

